PR #48, opened 2019-04-04
Fix creating match validation.

SeryMa:
When creating a match you could have created an invalid match (with "None" considered as a player).

One had to add a player and then remove him by selecting "None".

The problem was with filtering players.
playerId is string
INVALID_PLAYER_ID is number

comparing with triple comparator always resulted in false. So no "None" player was filtered out.

### DIFF
--- a/frontend/src/app/components/CreateMatch/SelectTeamForm.js
+++ b/frontend/src/app/components/CreateMatch/SelectTeamForm.js
@@ -33,7 +33,7 @@ export class SelectTeamForm extends Component {
                 selectedPlayerIds
             }
         }, () => {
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/CreateMatch/SelectTeamForm.js
```diff
@@ -33,7 +33,7 @@ export class SelectTeamForm extends Component {
                 selectedPlayerIds
             }
         }, () => {
-            const validPlayers = this.state.selectedPlayerIds.filter(playerId => playerId !== INVALID_PLAYER_ID)
+            const validPlayers = this.state.selectedPlayerIds.filter(playerId => playerId != INVALID_PLAYER_ID)
```
petrk-salsita: What is the reason for this change?

PR #49, opened 2019-04-11
Add foosbot

jlelic:


backend/bot/index.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  winningTeam.forEach(player => {
+    if (player.name !== 'Pepa' && player.name !== 'Tonda') {
+      isComedyDuo = false;
+    }
+  })
+
+  let prefix = '';
+  if (isComedyDuo) {
+    prefix = ':tondab: :pepadab: ';
```
petrk-salsita: Move the trailing space to the messageText? (so that it doesn't need to be repeated)

backend/bot/index.js
```diff
@@ -0,0 +1,60 @@
...
+  } else if (ratingChange < 10) {
+    prefix = 'Easy. ';
```
petrk-salsita: Same

backend/bot/index.js
```diff
@@ -0,0 +1,52 @@
...
+  } else if (ratingChange > 20) {
+    prefix = 'HOLY SHIT!';
+  }
+
+  let suffix = '';
```
petrk-salsita: Use `const` and single expression

backend/repositories/match-repository.js
```diff
@@ -79,5 +80,6 @@ const constructMatch = async (matchDescription) => {
   */
 exports.recordMatch = async (matchDescription) => {
     const match = await constructMatch(matchDescription)
+    bot.postResultToSlack(match);
```
petrk-salsita: Wait until the match is stored (there can be exceptions thrown) and post afterwards.

backend/bot/index.js
jlelic: Then there would be empty space at the start of the `messageText` if the prefix ends up being empty.

petrk-salsita: Makes sense... then add it to 'HOLY SHIT!', where it's currently missing please 😄

backend/bot/index.js
```diff
@@ -0,0 +1,60 @@
+var SlackBot = require('slackbots');
+
+var bot = new SlackBot({
+  token: process.env.FOOSBOT_TOKEN
+});
+
+let channel;
+
+console.log('lez go');
```
novellizator: delete comment

backend/bot/index.js
petrk-salsita: Oh, it's already there... nevermind!

PR #51, opened 2019-04-24
[BOT] Send ranking change to Slack channel and update its purpose

jlelic:


backend/bot/index.js
```diff
@@ -14,7 +29,23 @@ bot.on('start', async () => {
   }
 });
 
-const postResultToSlack = (match) => {
+const updateSlack = async (match) => {
```
novellizator: this function is definitely too long and the resopnsilibities are the following:
1. update channel purpose
2. print slack message, augmented if comedy duo plays 

pls break it down

backend/bot/index.js
```diff
@@ -14,7 +29,23 @@ bot.on('start', async () => {
...
+  // Update channel purpose
+  const users = await storage.getAllUsers(true);
+  const purposeCharacterLimit = 250;
+  let newChannelPurpose = '';
+  let i = 1;
+  for(const user of users) {
```
novellizator: learn to program my fren 
if users is an array, then array.forEach(), otherwise
is 
Object.entries(users).forEach()
both look better than having a for loop and a while loop at one place

petrk-salsita: ugggh, ugly!

backend/storage/storage-context.js
```diff
@@ -10,10 +10,12 @@ class StorageContext {
         this.transaction = transaction
     }
 
-    async getAllUsers() {
+    async getAllUsers(sortByRating = false) {
         let rows
         try {
-            rows = await this.transaction.executeQuery(dbQueries.selectAllusers, [])
+            rows = await this.transaction.executeQuery(
```
novellizator: this is problematic. what if we'll have to have the users sorted by name? or by  registration date? are we intending to add arguments in this function?

sorting users on client will never be problematic, because the number of them will always be super small.

what i suggest:
either getAllUsers().sort(...)
or getUsersSortedByRating()

backend/bot/index.js
```diff
@@ -14,7 +29,23 @@ bot.on('start', async () => {
   }
 });
 
-const postResultToSlack = (match) => {
+const updateSlack = async (match) => {
+  // Update channel purpose
```
novellizator: upon separating the responsibilities through function extraction, you can drop this comment

backend/bot/index.js
```diff
@@ -14,7 +29,23 @@ bot.on('start', async () => {
...
+    const purposeLine = `${i}. ${user.name} ${user.rating}\n`;
+    if(purposeLine.length + newChannelPurpose.length > purposeCharacterLimit) {
+      break;
+    }
+    newChannelPurpose += purposeLine;
+    i++;
+  }
+  await bot.setGroupPurpose(channel.id, newChannelPurpose)
+
+  // Send match info message
```
novellizator: and drop this one as well

backend/bot/index.js
```diff
@@ -1,6 +1,21 @@
-var SlackBot = require('slackbots');
+const SlackBot = require('slackbots');
 
-var bot = new SlackBot({
+const storage = require('../storage/storage');
```
petrk-salsita: Plz don't access `storage` module directly from here, fetch players from the `match-repository` in the `recordMatch` method and pass it to the `bot` module in a method argument.

backend/bot/index.js
```diff
@@ -14,7 +29,23 @@ bot.on('start', async () => {
...
-const postResultToSlack = (match) => {
+const updateSlack = async (match) => {
+  // Update channel purpose
+  const users = await storage.getAllUsers(true);
+  const purposeCharacterLimit = 250;
+  let newChannelPurpose = '';
+  let i = 1;
+  for(const user of users) {
+    const purposeLine = `${i}. ${user.name} ${user.rating}\n`;
+    if(purposeLine.length + newChannelPurpose.length > purposeCharacterLimit) {
```
novellizator: why not empty space between "in" and "(" ???

petrk-salsita: *`if`

backend/bot/index.js
```diff
@@ -36,12 +62,48 @@ const postResultToSlack = (match) => {
...
 
-  const suffix = isComedyDuo?  ':marioluigi:' : '';
+  const suffix = isComedyDuo ? ':marioluigi:' : '';
 
   const messageText = `${prefix}${winningPlayers.join(', ')} just beat ${losingPlayers.join(', ')}. Rating change: ${ratingChange}${suffix}`
-  bot.postMessage(channel.id, messageText, {as_user: true});
+  await bot.postMessage(channel.id, messageText, { as_user: true });
+}
+
+const postRankingChangeMessage = async (oldRankings, newRankings) => {
```
novellizator: i'd actually split this function, because this does two things.
1. form a message listing out the players
2. check if the purpose should be updated

Too much magic 🌈.

backend/bot/index.js
```diff
@@ -14,7 +27,20 @@ bot.on('start', async () => {
...
 
-const postResultToSlack = (match) => {
+
+const ratingComparator = (a, b) => b.rating - a.rating
+
+const reportMatchOnSlack = async (match, oldUsers, newUsers) => {
+  await postMatchResult(match)
+  const oldRankings = oldUsers.sort(ratingComparator)
+  const newRankings = newUsers.sort(ratingComparator)
+  const shouldUpdatePurpose = await postRankingChangeMessage(oldRankings, newRankings)
```
novellizator: await postRankingChangeMessage(oldRankings, newRankings)
const shouldUpdatePurpose = hasLeaderboardChanged(oldRankings, newRankings)
if (shouldUpdatePurpose) {
  await updtePurpose(...)
}

backend/bot/index.js
```diff
@@ -36,12 +62,48 @@ const postResultToSlack = (match) => {
...
+    )
+    .join('\n')
+  await bot.postMessage(channel.id, messageText, { as_user: true });
+
+  return shouldUpdatePurpose
 }
 
+const updatePurpose = async (rankings) => {
+  const purpose = ['TOP PLAYERS:']
+  for (let i = 0; i < 5; i++) {
```
novellizator: this `< 5` is tightly coupled with "<= 4" in `postRankingChangeMessage`, yet it's in a different function

novellizator: i'd make updatePurpose dumb printing out the whole array provided

backend/bot/index.js
```diff
@@ -36,12 +62,48 @@ const postResultToSlack = (match) => {
...
+  const rankingChanges = []
+  oldRankings.forEach((oldPlayer, index) => {
+    const newPlayer = newRankings[index]
+    if (newPlayer.id !== oldPlayer.id) {
+      if (index <= 4) {
+        shouldUpdatePurpose = true
+      }
+      rankingChanges.push({
+        name: oldPlayer.name,
+        oldRanking: index +1 ,
```
novellizator: spacing + 1

backend/bot/index.js
novellizator: maybe a separate class - leaderboardManager(oldrankoings, newrankings)
with functions
- hasLeaderBoardChanged
- getLeaderBoardRankings
?
this would separate the leaderboard related logic 
let me think

jlelic: I can't tell if you're serious rn or not

backend/bot/index.js
```diff
@@ -36,12 +68,61 @@ const postResultToSlack = (match) => {
...
-  bot.postMessage(channel.id, messageText, {as_user: true});
+  await bot.postMessage(channel.id, messageText, { as_user: true });
 }
 
+const postRankingChangeMessage = async (oldRankings, newRankings) => {
+  const rankingChanges = []
+  oldRankings.forEach((oldPlayer, index) => {
+    const newPlayer = newRankings[index]
+    if (newPlayer.id !== oldPlayer.id) {
+      if (index <= 4) {
```
novellizator: pls remove this if(index<=4) as well and put in a separate function. or at the very least use maxSignificantIndex here

jlelic: I've already created a separate function, I just forgot to remove this piece of code

backend/bot/index.js
```diff
@@ -36,12 +68,58 @@ const postResultToSlack = (match) => {
...
+  const suffix = isComedyDuo ? ':marioluigi:' : '';
 
   const messageText = `${prefix}${winningPlayers.join(', ')} just beat ${losingPlayers.join(', ')}. Rating change: ${ratingChange}${suffix}`
-  bot.postMessage(channel.id, messageText, {as_user: true});
+  await bot.postMessage(channel.id, messageText, { as_user: true });
 }
 
+const postRankingChangeMessage = async (oldRankings, newRankings) => {
+  const rankingChanges = []
+  oldRankings.forEach((oldPlayer, index) => {
```
petrk-salsita: Please use map with index argument instead of pushing to a mutable array.

backend/bot/index.js
```diff
@@ -36,12 +68,58 @@ const postResultToSlack = (match) => {
...
+
+  const messageText = rankingChanges
+    .map((c) => (
+      `${c.name} ${c.oldRanking}. ⟶ ${c.newRanking}.`)
+    )
+    .join('\n')
+  await bot.postMessage(channel.id, messageText, { as_user: true });
+}
+
+const hasLeaderboardChanged = (maxSignificant, oldRankings, newRankings) => {
```
petrk-salsita: `maxSignificant` -> `leaderboardSize` or something similar.

backend/bot/index.js
```diff
@@ -36,12 +68,58 @@ const postResultToSlack = (match) => {
...
+  const clampedMax = Math.min(maxSignificant, Math.min(oldRankings.length, newRankings.length))
+
+  for (let i = 0; i < clampedMax; i++) {
+    if (oldRankings[i].id != newRankings[i].id) {
+      return true
+    }
+  }
+
+  return false
```
petrk-salsita: If JS had zip, we could write
```
return oldRankings.take(clampedMax)
  .zip(newRankings.take(clampedMax))
  .any((old, new) => old.id != new.id)
```
😢

backend/bot/index.js
jlelic: How would this work if I'm only pushing when `newPlayer.id !== oldPlayer.id`? 🤔

Should I just `map` it and then use `filter`?

jlelic: or like this?
```
const hasLeaderboardChanged = (leaderboardSize, oldRankings, newRankings) => (
  oldRankings.findIndex((oldPlayer, index) => oldPlayer.id !== newRankings[index].id) < leaderboardSize
```

backend/bot/index.js
jlelic: or we can do this? 🤔
```
hasLeaderboardChanged = (leaderboardSize, oldRankings, newRankings) => (
  oldRankings.findIndex((oldPlayer, index) => oldPlayer.id !== newRankings[index].id) < leaderboardSize
)
```

backend/bot/index.js
petrk-salsita: You can do it like this:
```
const rankingChanges = oldRankings.map((oldPlayer, index) => ({
  name: oldPlayer.name,
  oldRanking: index + 1,
  newRanking: newRankings.findIndex(p => p.id === oldPlayer.id) + 1
})).filter(ranking => ranking.oldRanking != ranking.newRanking)
```

backend/bot/index.js
petrk-salsita: That's nice as well! 😸

backend/bot/index.js
jlelic: Yeah that's what I meant, I just wasn't sure if it's what you meant too

jlelic: Actually I think this way there may be a problem if we had several people with the same rating, since we are only filtering by comparing ratings.

petrk-salsita: Ranking is the order - you can't have 2 ppl with the same one :D

jlelic: Exactly, I deleted my previous comment immediately after posting it because I realized that.

petrk-salsita:
You did an amazing job sir! Please feel free to merge whenever ready!

jlelic:
Thanks guys, it wouldn't have been possible without you 🙌

PR #52, opened 2019-06-18
Adjust the prefix thresholds for the Slack bot.

petrk-salsita:


petrk-salsita:
@jlelic Thank you for your valuable and much appreciated input. I'm afraid, however, that I'm not gonna do it. Nobody is asking you to review this anyway 😉(except for me in the channel, but that doesn't count...)

PR #53, opened 2019-07-16
Fix prefix

padronhosnai:


backend/bot/index.js
```diff
@@ -33,7 +33,7 @@ const postResultToSlack = (match) => {
   } else if (ratingChange <= 10) {
     prefix = 'Easy. ';
   } else if (ratingChange >= 20) {
-    prefix = 'HOLY SHIT! ';
+    prefix = 'HOLY SHIT! L2P, noob! ';
```
jlelic: We should probably use `noobs` in case there are  players in the losing team

backend/bot/index.js
```diff
@@ -33,7 +33,7 @@ const postResultToSlack = (match) => {
   } else if (ratingChange <= 10) {
     prefix = 'Easy. ';
   } else if (ratingChange >= 20) {
-    prefix = 'HOLY SHIT! ';
+    prefix = 'HOLY SHIT! L2P, noobs! ';
```
jlelic: There can still be only one player in the losing team though

padronhosnai:
Oh, sorry. I thought this was blockchain based and that I would get a commit reward for every commit. My bad.

PR #54, opened 2019-07-23
Anti-Tonda protection cooldown

petrk-salsita:


backend/repositories/match-repository.js
```diff
@@ -72,13 +73,26 @@ const constructMatch = async (matchDescription) => {
     return new Match(teams, matchDescription.team1Won, date, ratingChange)
 }
 
+const getElapsedSecondsSinceLastMatch = async () => {
+    const matchTimes = (await storage.getAllMatches()).map(match => match.date.getTime())
```
jlelic: all matches? 😱

petrk-salsita: Is good enough :smile

zbrcz: How hard would it be to implement getting only the last match from the DB?

zbrcz:
I like the name of the PR. :D

backend/repositories/match-repository.js
```diff
@@ -3,6 +3,7 @@ const ratingCalculator = require("../rating/rating-calculator")
 const { InputError } = require('../errors/input-error')
 const { NotFoundError } = require('../errors/not-found-error')
 const bot = require('../bot');
+const ADD_MATCH_COOLDOWN = 60
```
jlelic: On a second thought, is 60 second enough? 🤔
There can be a scenario when user tries to add a match while the server on heroku is sleeping. The request will still be sent and heroku will wait for the server to wake up to process it, but the user in the meanwhile can refresh the page, go grab a coffee yada yada, and when they come back the minute could have already elapsed and they add a duplicate match since they won't see that the first request actually came through.
I would go with 5 minutes just to be safe.

backend/repositories/match-repository.js
petrk-salsita: One SQL query, not interested in writing those though. I kinda like ORM 😄

backend/repositories/match-repository.js
petrk-salsita: Ok, I'll make it configurable.

backend/repositories/match-repository.js
petrk-salsita: Ok, fixed that! Please re-review

petrk-salsita:
I'm flattered! ❤️

jlelic:
That's not much of a compliment

PR #55, opened 2019-08-01
Protection against unwanted match additions, double additions, wrong result recordings, etc

novellizator:


backend/app.js
```diff
@@ -51,6 +51,7 @@ app.post('/users', (req, res) => {
 })
 
 app.post('/matches', (req, res) => {
+    if (req.connection.remoteAddress == "192.168.23.86") return
```
jlelic: > 192.168.23.86

Where do I even begin...

novellizator:
closing for now, hopefully we won't need it:)

PR #57, opened 2019-08-16
Fix rating change when playing 2v1

zbrcz:
@petrk-salsita Please add 10 ELO to Tonda when this is merged.

### DIFF
--- a/backend/repositories/match-repository.js
+++ b/backend/repositories/match-repository.js
@@ -22,8 +22,32 @@ const storeMatch = async (match) => {
     try {
         result = await storageContext.insertMatch(match)
 
-        await updateRatingForTeam(winningTeam, match.ratingChange, storageContext)
-        await updateRatingForTeam(losingTeam, -match.ratingChange, storageContext)
+        // FIXME: foosbot's report of rating change is incorrect if the number of players on each side is not the same
+        // one player beat two players
+        if (winningTeam.length < losingTeam.length) {
+            // he gains rating from both opponents
+            await updateRatingForTeam(winningTeam, 2 * match.ratingChange, storageContext)
+            // they each lose half of what he gains
+            await updateRatingForTeam(losingTeam, -match.ratingChange, storageContext)
+
+        // two players beat one player
+        } else if (winningTeam.length > losingTeam.length) {
+            // they split his rating in half (rounded up)
+            const splitRatingChange = match.ratingChange / 2
+            const actualRatingChange = Math.round(splitRatingChange)
+            await updateRatingForTeam(winningTeam, actualRatingChange, storageContext)
+            if (actualRatingChange !== splitRatingChange) {
+                // rounding gave winners extra half a point each, loser needs to lose one more
+                await updateRatingForTeam(losingTeam, -(match.ratingChange + 1), storageContext)
+            } else {
+                await updateRatingForTeam(losingTeam, -match.ratingChange, storageContext)
+            }
+
+        // two players beat two or one player beat one
+        } else {
+            await updateRatingForTeam(winningTeam, match.ratingChange, storageContext)
+            await updateRatingForTeam(losingTeam, -match.ratingChange, storageContext)
+        }
 
         await storageContext.commit()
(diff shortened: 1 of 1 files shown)


jlelic:
You forgot to ask the product owner for a review 🤦‍♂️

backend/repositories/match-repository.js
```diff
@@ -22,8 +22,32 @@ const storeMatch = async (match) => {
     try {
         result = await storageContext.insertMatch(match)
 
-        await updateRatingForTeam(winningTeam, match.ratingChange, storageContext)
-        await updateRatingForTeam(losingTeam, -match.ratingChange, storageContext)
+        // FIXME: foosbot's report of rating change is incorrect if the number of players on each side is not the same
```
novellizator: fixme comments don't belong in production code. pls remove comment and create an issue for that

backend/repositories/match-repository.js
```diff
@@ -22,8 +22,32 @@ const storeMatch = async (match) => {
...
+        // one player beat two players
+        if (winningTeam.length < losingTeam.length) {
```
novellizator: This function is called storeMatch and its only responsibility is to store the match, not to do some computing logic. i'd edit edit the logic of whatever generates the match to do this.

maybe augment the match with two properties:
{team1RatingChange:..., team2RatingChange:...}

backend/repositories/match-repository.js
```diff
@@ -22,8 +22,32 @@ const storeMatch = async (match) => {
...
+            // he gains rating from both opponents
```
jlelic: Please refrain from using gender specific pronouns when referring to users/players. Either use `a player gains` or https://en.wikipedia.org/wiki/Singular_they

backend/repositories/match-repository.js
```diff
@@ -22,8 +22,32 @@ const storeMatch = async (match) => {
     try {
         result = await storageContext.insertMatch(match)
 
-        await updateRatingForTeam(winningTeam, match.ratingChange, storageContext)
-        await updateRatingForTeam(losingTeam, -match.ratingChange, storageContext)
+        // FIXME: foosbot's report of rating change is incorrect if the number of players on each side is not the same
+        // one player beat two players
```
jlelic: The condition says `winningTeam.length < losingTeam.length` which doesn't necessarily mean that `one player beat two players`. I would either make the condition more specific or change the calculation to take into the account different numbers of players in the team to save us potential trouble down the line.

zbrcz:
@novellizator Yeah, I found out over the weekend that you'll probably take it way more seriously than I had expected... Ok, I'll change the implementation, not sure when I'll have the time though...

backend/repositories/match-repository.js
novellizator: good catch!

petrk-salsita: Great catch indeed, I can't imagine what could have happened if something this serious slipped into our production code 😱.

backend/repositories/match-repository.js
zbrcz: Based on the documentation it does. Take a look at this:
```
  * @param {Array<number>} matchDescription.team1 Array of 1 or 2 elements containing IDs of players from team1.
  * @param {Array<number>} matchDescription.team2 Array of 1 or 2 elements containing IDs of players from team2.
```
https://github.com/salsita/foosball-rating/blame/master/backend/repositories/match-repository.js#L89
But sure, you are right. I knew about it as I was implementing it but I really didn't expect everyone to suddenly start taking it so seriously... As I said above, not sure when I'll have the time (and the will) to change all this...

backend/repositories/match-repository.js
zbrcz: Yeah, I purposefuly decided to go with a commonly used error in order not to use "they" for both the two players and the single one. Thought it would be more clear. And once again, I didn't think you'd take it this seriously. But sure, I'll change it. One day...

zbrcz:
Will re-implement with all seriousness at some point...

PR #58, opened 2019-08-16
Rating history graps in user profile

jlelic:
- Added rating history graph to user profile.
- Rewrote the two components I touched to functional components. It's not exactly consistent with the rest of the application but it's better ¯\\_(ツ)_/¯


frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -15,6 +15,23 @@ export const generateMatchRatingChanges = (userId, userMatches) => userMatches.m
     }
 })
 
+export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => userMatches
```
jlelic: I think `generateRatingHistoryGraphForUser` may be too long 🤔

petrk-salsita: I think it's perfectly fine and descriptive

frontend/src/app/pages/Profile/index.js
```diff
@@ -1,30 +1,37 @@
...
   render() {
     if (!this.props.user) {
       return <div />
     }
 
-    return(
+    const {
+      user: { name, rating, id },
+      statistics: { totalMatches, winRatio, longestStreak, matchChanges }
+    } = this.props
```
petrk-salsita: Some of the variables from the deconstruction are not used, right?

petrk-salsita:
What an amazing piece of software architecture, thank you!

frontend/src/app/pages/Profile/index.js
jlelic: Nope, all of them are used.

jlelic:
It's not every day you get to see a stateful `map` :genius:

PR #60, opened 2019-08-19
Fix rating change when playing 2v1, version 2

zbrcz:
To do:
1. Run the SQL in production before pushing it there.
2. Give Tonda 10 ELO.


zbrcz:
The whole point of this PR is so that it does. Because currently, it doesn't.

backend/bot/index.js
```diff
@@ -59,19 +59,28 @@ const postMatchResult = async (match) => {
...
-  let prefix = '';
+  const messageParts = []
+
   if (isComedyDuo) {
-    prefix = ':tondab: :pepadab: ';
-  } else if (ratingChange <= 10) {
-    prefix = 'Easy. ';
-  } else if (ratingChange >= 20) {
-    prefix = 'HOLY SHIT! ';
+    messageParts.push(':tondab: :pepadab:')
```
novellizator: i know it's outside of the scope of the PR, bt yo touched the file, so I'm looking at you! 
now that we're opensourcing this project, could you please add a comment in the sources that explains the comedy duo?

backend/db_migrations/winning-losing-rating-change.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE "Matches" RENAME COLUMN "RatingChange" TO "WinningTeamRatingChange";
```
novellizator: i don't know how we're running the migrations, BUT
1. don't we want to order the migration file names?

0-setup-database.sql
1.-winning-losing-rating-change.sql
2.--- 
?

2. how about mentioning it in the readme?

backend/bot/index.js
novellizator: theoretically, you could have the messageParts built like you do and then do somthing like

applyComedyDuoMiddleware(messageParts, match) to make a pseudo chain-or responsibility thing. You don't need to do this if you don't want to though.

zbrcz: I can add it but tell me what the comment should be because I have no idea how to explain that. :D Or just add the comment yourself then.

Sorry, I don't have the time to even think about the second part. :D

backend/db_migrations/winning-losing-rating-change.sql
zbrcz: I think we should order them at the very least. I can rename the files.

What exactly do you want to be mentioned in the readme?

novellizator: anything that would help the uninitiated run the app. just mention it in the setup phase, that they should run the migrations

backend/bot/index.js
novellizator: "Comedy duo refers of two salsitistas known for their distinctive playing style" :D

zbrcz: I will create a separate PR to generalize the functionality. See Slack.

zbrcz: I have the code with the generalized "comedy duo" functionality ready but it's dependent on the code from here. Will create a new PR once this one is merged.

zbrcz:
@novellizator Please merge this when you're ready to run the migration at the same time. Then please give Tonda 10 ELO or replay all the matches. Thanks.

PR #61, opened 2019-08-29
Add information about environment variables

novellizator:


backend/README.md
```diff
@@ -30,3 +31,19 @@ npm start
 ```
 
 The port on which the server is listening is reported in the console output.
+
+## Environment variables
+Here's the list of the envirnment variables that could to be configured.
```
zbrcz: 1) "environment" instead of "envirnment".
2) "could be configured" instead of "could to be configured".
3) Personally, I'd add an empty line above this one.

backend/README.md
```diff
@@ -30,3 +31,19 @@ npm start
...
+```
+process.env.DATABASE_URL
+```
+- stores the url to the database. If the variable is missing, the backend connects to our test database. 
+
+```
+process.env.FOOSBOT_TOKEN
+process.env.FOOS_CHANNEL_NAME
+```
+- these variables are used by the slackbot to announce the results of the matches You can obtain the bot Token on slack settings page and then select the channel you'd like to deploy the bot on.
```
zbrcz: You're missing a period after "matches".

backend/README.md
novellizator: fixed, thanks!

backend/README.md
novellizator: fixed

PR #62, opened 2019-08-29
readme: add a note on running the app locally

novellizator:


novellizator:
yeah, that needs to be changed as well.

PR #63, opened 2019-08-30
Generalize comedy duo functionality

zbrcz:


backend/bot/index.js
```diff
@@ -57,12 +62,12 @@ const postMatchResult = async (match) => {
   const winningPlayers = winningTeam.map(player => `${player.name} (${player.rating})`)
   const losingPlayers = losingTeam.map(player => `${player.name} (${player.rating})`)
 
-  const isComedyDuo = winningPlayers.length == 2 && winningTeam.every(player => player.name === 'Pepa' || player.name === 'Tonda')
-
   const messageParts = []
 
-  if (isComedyDuo) {
-    messageParts.push(':tondab: :pepadab:')
+  const { prefix, suffix } = winningTeam.length === 2 ? getMatchResultPrefixSuffix(winningTeam) : {}
```
novellizator: why is it limited to 2 people?

zbrcz: Frankly, I didn't have the time to deal with that option. It can be added later.

backend/app.js
```diff
@@ -24,7 +24,11 @@ app.use(urlencodedParser)
 app.use(jsonParser)
 
 let bot 
-botFactory.makeBot(process.env.FOOSBOT_TOKEN, process.env.FOOS_CHANNEL_NAME)
+botFactory.makeBot(
```
novellizator: prefix suffix assignment is not a part of the bot's respobnsibility. bot should only be able to post messages to a channel. it's the responsibility of the match reporter (different file). please don't put it in the bot

zbrcz:
Reworked the match repository and reporter a bit - I think it makes much more sense like this.

backend/app.js
```diff
@@ -31,7 +32,7 @@ botFactory.makeBot(process.env.FOOSBOT_TOKEN, process.env.FOOS_CHANNEL_NAME)
     })
     .catch((error) => console.log('Bot initialization failed', error))
 
-
+const matchReporter = new MatchReporter(process.env.MATCH_RESULT_PREFIX_SUFFIX_CONFIG)
```
novellizator: thinking - what if match reporter also took the bot as a constructor?

backend/bot/parseMatchResultPrefixSuffixConfig.js
```diff
@@ -0,0 +1,16 @@
+const parseConfigPart = configPart => {
+  const splitConfigPart = configPart.split(',')
+  if (splitConfigPart.length !== 4) {
+    throw Error(`Could not parse config part "${splitConfigPart}"`)
+  }
+  return ({
```
novellizator: 1. wondering, why the extra parentheses in return ({}) ? and not just return {} ? It it's a normal JS style, keep it like it is ;)
2. i'd probably need to mull this over, but it seems that this function is tightly coupled with the match reporter, so would consider putting it in the same file. alternatively would at least consider renaming this loong funciton name to parseMatchReportDecorables(...) or somethign such

backend/repositories/match-repository.js
```diff
@@ -91,26 +90,15 @@ const getElapsedSecondsSinceLatestMatch = async () => {
...
-    const newUsers = await storage.getAllUsers()
-
-    if (bot) {
-        try {
-            await matchReporter.reportMatchOnSlack(bot, match, oldUsers, newUsers)
-        } catch (error) {
-            console.log('Bot error', error)
-        }
-    }
+    await storeMatch(match)
```
novellizator: i like this refactoring:)

backend/bot/MatchReporter.js
```diff
@@ -47,13 +60,21 @@ const createMatchResultMessage = (match) => {
...
-  if (isComedyDuo) {
-    messageParts.push(':marioluigi:')
+  if (suffix) {
+    messageParts.push(suffix)
   }
 
   return messageParts.join(' ')
 }
 
+const getMatchResultPrefixSuffix = (prefixSuffixConfig, winningTeam) => {
```
novellizator: findDecoratorForTeam(decorationConfig, team) ? 🤷‍♂ 

also then, should return a falsy thing...

backend/bot/MatchReporter.js
```diff
@@ -1,23 +1,34 @@
...
+
+  async reportMatchOnSlack(bot, match, oldUsers, newUsers) {
+    const matchResultMessage = createMatchResultMessage(match, this.prefixSuffixConfig)
+
+    const oldRankings = oldUsers.sort(ratingComparator)
+    const newRankings = newUsers.sort(ratingComparator)
+
+    const rankingChangeMessage = createRankingChangeMessage(oldRankings, newRankings)
+    await bot.postMessage(`${matchResultMessage}\n${rankingChangeMessage}`)
+    const leaderboardSize = 5
```
novellizator: probably put into construcotr, maybe with a default value of 5...

backend/app.js
zbrcz: `bot` is still `undefined` at the time the constructor is called, so it's not that simple. But what you asked for makes perfect sense... So I am now passing all the env. variables to the constructor and initializing the bot there. Plus I took the refactoring even furhter... I'm gonna push it soon. :)

backend/bot/parseMatchResultPrefixSuffixConfig.js
zbrcz: 1. http://jamesknelson.com/javascript-return-parenthesis/
2. Gonna put it within the match reporter.

backend/README.md
```diff
@@ -48,4 +48,14 @@ process.env.FOOS_CHANNEL_NAME
 ```
 - these variables are used by the slackbot to announce the results of the matches. You can obtain the bot Token on slack settings page and then select the channel you'd like to deploy the bot on. 
 
-
+```
+process.env.MATCH_RESULT_PREFIX_SUFFIX_CONFIG
+```
+- used to configure prefixes and suffixes of the matches results messages (see above) based on the players on the winning team.
+- the app expects quadruplets of player1, player2, prefix and suffix to be encoded in the environment variable.
+- use `,` to separate the values within a quadruplet (`player1,player2,prefix,suffix`) and `;` to separate the quadruplets (`quadruplet1;quadruplet2`). Note: the order of players in the quadruplet does not matter.
```
novellizator: also, i'd consider making it more generic - like "<numberOfRequeredPlayers>, <names>, prefix, sufix", eg "2Pepa, Tomy,Stego,michalvenos,good idea,slovakia" "1,Pepa,Tomy,Stego, heyhey, one slovak guy on a team"

this might be a bit too much though.... :)

backend/bot/MatchReporter.js
zbrcz: I like the whole `decorator` and `decoration` idea. Gonna go with it.

backend/bot/MatchReporter.js
zbrcz: Not my code, but sure.

backend/README.md
zbrcz: Sorry, I don't have the time for that anymore.

backend/match-reporter/match-reporter.js
```diff
@@ -0,0 +1,141 @@
+const botFactory = require('./bot-factory');
+
+const DEFAULT_LEADERBOARD_SIZE = 5;
+
+class MatchReporter {
+  constructor(botToken, channelName, matchReportPrefixSuffixConfig, leaderboardSize = DEFAULT_LEADERBOARD_SIZE) {
```
novellizator: i'd globally rename the `prefixsuffix` to decorator or soemthing such

backend/app.js
```diff
@@ -23,15 +24,11 @@ app.use(addCrossDomainHeaders)
...
-botFactory.makeBot(process.env.FOOSBOT_TOKEN, process.env.FOOS_CHANNEL_NAME)
-    .then(resolvedBot => {
-        bot = resolvedBot
-        console.log('Slackbot initialized!')
-    })
-    .catch((error) => console.log('Bot initialization failed', error))
-
-
+const matchReporter = new MatchReporter(
+  process.env.FOOSBOT_TOKEN,
```
novellizator: i'd like to think of a bot as a separate entity. potentially used for recording a match, responding to various stats upon request etc. - it shuold be initialized within the match reporter but separtely

backend/match-reporter/bot-factory.js
```diff
@@ -17,24 +17,27 @@ class SingleChannelBot {
...
+
+    slackbot.on('error', (error) => reject(error.message))
+
     slackbot.on('start', async () => {
       let groupId 
       try {
         groupId = await slackbot.getGroupId(channelName)
       } catch (error) {
-        return reject(error)
+        reject(error.message)
```
novellizator: 1. why are you rejecting with a string? i thought it's better to reject with a stack trace. more so since you're catching it later on (throwing non Error types is an antipattern)
2. why did you remove the return from revolse and reject? i couldn't find it after a quick search, but is it in the spec, what happens  if you call one of the functions multiple times?

backend/README.md
novellizator: I AM THE CODE OWNER, OBEY ME

backend/match-reporter/bot-factory.js
zbrcz: 1. Well, when you reject, you don't really throw... But sure, we can reject with `Error`s.
2. My bad. I'll fix it. Though the `return` doesn't make much sense there IMO, so I guess I'll fix it by resolving within `try` rather than returning `return` back. ;)

backend/README.md
novellizator: JK ;)

novellizator:
1.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject#Description
whether
you throw it or not is irrelevant, according to this, rejected reason
should be an error

backend/match-reporter/bot-factory.js
novellizator: 1.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject#Description
whether
you throw it or not is irrelevant, according to this, rejected reason
should be an error

zbrcz: That's not what the documentation says. It says it may be useful for it to be an instance of `Error`, not that it has to be that. Anyway, I changed it.

pilartomas:
```code
const makeBot = async (botToken, channelName) => {
	if(!botToken || !channelName) throw new Error('botToken or channelName missing');
	const slackbot = new SlackBot({ token: botToken });
	return Promise.race([
		new Promise((resolve, reject) => slackbot.on('error', error => reject(error)),
		new Promise((resolve, reject) => slackbot.on('start', () =>
                   slackbot.getGroupId(channelName)
                   .then(groupId => resolve(new SingleChannelBot(slackbot, channelName, groupId)))
                   .catch(error => reject(error))
		)
	]);
}
```
`new Promise()` only the callbacks dammit :dagger: :dagger: :D

zbrcz: @pilartomas This would mean that you'd have to catch errors in the promise chain as well as surround calling `makeBot` with a `try/catch` block to catch the error you throw on the first line. I think it's better to handle the errors only in one place. Anyway, I like that `race` thing. Feel free to create a PR with that change. ;) I am done here. :D

pilartomas: @zbrcz Nope for the first sentence. Throwing from `async` function (I made `makeBot` async) gets compiled into rejection. The code should work same as the old one from the caller's viewpoint :)

zbrcz: Ah, I didn't notice that you made it async... Anyway, as I said, feel free to create a PR. ;)

PR #64, opened 2019-08-30
Add initial rating to user graphs

zbrcz:
In order for the graph not to start with a completely upwards or downwards facing line, we assign the initial rating to the user one day before their first match. It's somehow as if the user has been added to the DB one day before their first match.

### DIFF
--- a/frontend/src/app/modules/matches/matches-computations.js
+++ b/frontend/src/app/modules/matches/matches-computations.js
@@ -17,9 +17,9 @@ export const generateMatchRatingChanges = (userId, userMatches) => userMatches.m
     }
 })
 
-export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => {
+export const plotRatingHistory = (userMatches, userId, initialRating) => {
   const reversedUserMatches = userMatches.reverse()
-  const firstMatchDate = userMatches[0] ? userMatches[0].date : null
+  const firstMatchDate = userMatches[0] && userMatches[0].date
   const graphStartDate = firstMatchDate ? new Date(firstMatchDate) : new Date()
   if (firstMatchDate) {
     graphStartDate.setDate(firstMatchDate.getDate() - 1)
--- a/frontend/src/app/modules/matches/matches-computations.js
+++ b/frontend/src/app/modules/matches/matches-computations.js
@@ -17,9 +17,15 @@ export const generateMatchRatingChanges = (userId, userMatches) => userMatches.m
     }
 })
 
-export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => userMatches
-  .reverse()
-  .map(
+export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => {
+  const reversedUserMatches = userMatches.reverse()
+  const firstMatchDate = userMatches[0] ? userMatches[0].date : null
+  const graphStartDate = firstMatchDate ? new Date(firstMatchDate) : new Date()
+  if (firstMatchDate) {
+    graphStartDate.setDate(firstMatchDate.getDate() - 1)
+  }
+  return [{ x: graphStartDate, y: initialRating }]
+    .concat(reversedUserMatches.map(
       function(match) {
         const didWin = didUserWin(userId, match)
         const change = didWin ? match.winningTeamRatingChange : match.losingTeamRatingChange
@@ -30,9 +36,10 @@ export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRa
           y: this.rating,
           x: match.date
         }
-    },
-    { rating: initialRating }
-  )
+      },
+      { rating: initialRating }
+    ))
+}
 
 export const computeLongestWinStreak = (userId, userMatches) => {
     const initialState = {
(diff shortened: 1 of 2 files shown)


frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -17,9 +17,10 @@ export const generateMatchRatingChanges = (userId, userMatches) => userMatches.m
     }
 })
 
-export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => userMatches
-  .reverse()
-  .map(
+export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => {
+  const reversedUserMatches = userMatches.reverse()
+  return [{ x: userMatches[0] ? new Date().setDate(userMatches[0].date.getDate() - 1) : new Date(), y: initialRating }]
```
jlelic: I'm assuming you want the initial rating to be on the graph one day before the first match, however I'm not sure the expression:
`new Date().setDate(userMatches[0].date.getDate() - 1)` 
would work in this case. You're just creating an object with current date (`new Date`) then setting the `date` which is actually just the day of the month (yaay javascript), so if the `userMatches[0].date` is say on `2018-6-6` the resulting initial date (today) would be `2019-9-5` and not `2018-6-5`.

zbrcz: Thanks, you pointed me to some minor issues with the code that I'll fix, but the thing that you point out works as expected:
```
> const today = new Date();
undefined
> today;
2019-09-02T09:40:24.694Z
> const yesterdayTs = (new Date()).setDate(today.getDate() - 1);
undefined
> const yesterday = new Date(yesterdayTs);
undefined
> yesterday;
2019-09-01T09:40:28.670Z
> const lastDayOfAugustTs = (new Date()).setDate(yesterday.getDate() - 1);
undefined
> const lastDayOfAugust = new Date(lastDayOfAugustTs);
undefined
> lastDayOfAugust;
2019-08-31T09:41:05.698Z
```

zbrcz: You're right actually, the method works, except I am using it wrong... :D

jlelic: No worries man, happens to the best of us

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -17,9 +17,15 @@ export const generateMatchRatingChanges = (userId, userMatches) => userMatches.m
...
-  .reverse()
-  .map(
+export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => {
+  const reversedUserMatches = userMatches.reverse()
+  const firstMatchDate = userMatches[0] ? userMatches[0].date : null
+  const graphStartDate = firstMatchDate ? new Date(firstMatchDate) : new Date()
+  if (firstMatchDate) {
+    graphStartDate.setDate(firstMatchDate.getDate() - 1)
+  }
+  return [{ x: graphStartDate, y: initialRating }]
```
novellizator: cannot we do it without concat? 

return [{ x: graphStartDate, y: initialRating }, ...evaluateMatchesInTime(reversedUserMatches)]

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -17,9 +17,15 @@ export const generateMatchRatingChanges = (userId, userMatches) => userMatches.m
     }
 })
 
-export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => userMatches
-  .reverse()
-  .map(
+export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => {
+  const reversedUserMatches = userMatches.reverse()
+  const firstMatchDate = userMatches[0] ? userMatches[0].date : null
```
novellizator: theoretically could be rewritten to const firstMatchDate = userMatches[0] && userMatches[0].date, 
but don't know how much this is just a matter of taste so feel free to leave it like it is

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -17,9 +17,15 @@ export const generateMatchRatingChanges = (userId, userMatches) => userMatches.m
     }
 })
 
-export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => userMatches
-  .reverse()
-  .map(
+export const generateRatingHistoryGraphForUser = (userMatches, userId, initialRating) => {
```
novellizator: plotRatingHistory() might sound better

frontend/src/app/modules/matches/matches-computations.js
zbrcz: I'd prefer not to. I believe the method with `concat` is faster. I don't like overusing spreading. However, if you put your foot down and say "I AM THE CODE OWNER, OBEY ME", I'll change it.

frontend/src/app/modules/matches/matches-computations.js
zbrcz: Not my code, but sure.

frontend/src/app/modules/matches/matches-computations.js
novellizator: leaving my foot up for on this one:)

zbrcz:
@jlelic Can you please lift the changes request?

PR #77, opened 2020-02-10
Frontend names bug fix

Fezzzi:


frontend/src/app/modules/users/users-effects.js
```diff
@@ -1,5 +1,16 @@
 import { client } from '../api/api-client'
 
+
+/**
```
novellizator: why do you have the User object defined twice? I suspect you want to remove one of the definitions....

frontend/src/app/modules/root/root-reducer.js
```diff
@@ -18,7 +18,10 @@ const initialState = {
 
 const usersLoaded = (state, { users }) => ({
   ...state,
-  users,
+  users: users.map((user) => {
+    user.id = Number(user.id)
```
novellizator: in general, mutating maps are a big no-no. with this map, you're editing the original user object.
```

users.map (user => {
  return {...user, id: Number(user.id)}
})
```

would be a way to go

frontend/src/app/modules/users/users-effects.js
novellizator: this JSDoc syntax only implies the existence of the User object with such and such properties - I believe it doesn't imply that getUsers() or addUser() return such an object.

also a typo duplicated by copy pasting the code (initilRating)

frontend/src/app/modules/root/root-reducer.js
novellizator: actually - yesterday we agreed not to fix it on backend due to the hanging PR. I was wrong. Consistent and good design is more important than efficiency. Could rather you fix it on the backend and ignoring the conflict in the hanging PRs? The authors of the hanging PRs should do the rebase

PR #78, opened 2020-02-11
WIP

novellizator:
do not emrge


novellizator:
closign, need to test the postdeploy script

PR #81, opened 2020-02-13
Issue 71 light theme

Fezzzi:


frontend/src/app/components/header.js
```diff
@@ -12,6 +12,15 @@ import { withRouter } from 'react-router-dom'
 const logo = require('./../../media/logo.png')
 
 class HeaderComponent extends Component {
+  changeTheme = () => {
+    this.props.changeTheme()
+
+    // Stop transition effects
+    window.setTimeout(() => {
```
petrk-salsita: This should ideally go to a saga so that the React component itself doesn't handle side effects. You'll probably want something like a saga that gets fired on every `CHANGE_THEME`, waits for the timeout period and then dispatches the `STOP_THEME_TRANSITION`. This also allows you to completely omit the `stopThemeTransition()` from `Header` props.

frontend/src/app/modules/root/root-reducer.js
```diff
@@ -14,6 +14,8 @@ const initialState = {
   users: [],
   activeAlert: null,
   activeRedirect: null,
+  theme: window.localStorage.getItem("theme") || "dark",
```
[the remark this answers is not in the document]
Fezzzi: If I remove it, the app will open each time with dark theme. This way it will remember theme selection before refresh. Is there any way around?

frontend/src/app/modules/root/root-reducer.js
```diff
@@ -86,11 +88,33 @@ const dismissAlert = state => {
   }
 }
 
+const changeTheme = state => {
+  const theme = state.theme === "dark" ? "light" : "dark"
```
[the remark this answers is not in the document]
Fezzzi: Totally agree

frontend/src/app/components/theme.js
```diff
@@ -0,0 +1,52 @@
...
+import { RootActions } from '../modules/root/root-actions'
+// pages
+import { Dashboard } from '../pages/Dashboard'
+import { CreateMatchPage } from '../pages/CreateMatch'
+import { Profile } from '../pages/Profile'
+import { AddUserPage } from '../pages/AddUser'
+import { MatchListPage } from '../pages/MatchList'
+import { UserListPage } from '../pages/UserList'
+
+const ThemeComponent = ({theme, themeTransition, changeTheme, stopThemeTransition}) => {
```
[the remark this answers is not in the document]
Fezzzi: agree, fixed

frontend/src/app/components/theme.js
```diff
@@ -0,0 +1,52 @@
...
+  const transition = themeTransition === 1 ? " theme-transition" : ""
+
+  return (
+    <Router>
+      <div className={`${theme}${transition}`}>
+        <Header changeTheme={changeTheme} stopThemeTransition={stopThemeTransition} />
```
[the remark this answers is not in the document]
Fezzzi: Agree, will fix it

frontend/src/app/components/theme.js
```diff
@@ -0,0 +1,52 @@
...
+import { AddUserPage } from '../pages/AddUser'
+import { MatchListPage } from '../pages/MatchList'
+import { UserListPage } from '../pages/UserList'
+
+const ThemeComponent = ({theme, themeTransition, changeTheme, stopThemeTransition}) => {
+  const transition = themeTransition === 1 ? " theme-transition" : ""
+
+  return (
+    <Router>
+      <div className={`${theme}${transition}`}>
```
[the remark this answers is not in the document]
Fezzzi: ClassNames looks great, thanks for the suggestion!

frontend/src/app/components/header.js
novellizator: you also don't need to write _window_ explicitly

frontend/src/app/components/theme.js
```diff
@@ -0,0 +1,52 @@
...
+// pages
+import { Dashboard } from '../pages/Dashboard'
+import { CreateMatchPage } from '../pages/CreateMatch'
+import { Profile } from '../pages/Profile'
+import { AddUserPage } from '../pages/AddUser'
+import { MatchListPage } from '../pages/MatchList'
+import { UserListPage } from '../pages/UserList'
+
+const ThemeComponent = ({theme, themeTransition, changeTheme, stopThemeTransition}) => {
+  const transition = themeTransition === 1 ? " theme-transition" : ""
```
novellizator: i ran the app on localhost and the linter seemed to be screaming all the time. in this instance about the double quotes. Do you have a linter installed?

frontend/src/app/modules/root/root-reducer.js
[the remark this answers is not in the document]
petrk-salsita: Makes sense, thanks. 🙂 I'd maybe reconsider the place where you set the theme - reducers should not have any side effects in principle, they should be pure transformation of previous state to a new state. The cleanest solution would probably be to have a saga that takes care of it 🤔

frontend/src/app/components/theme.js
Fezzzi: no, I have no linter installed. Are we talking about IDE extension I suppose? React linter pops up a lot of hits, which one shall I choose?

novellizator:
yes, you need an IDE extension. you have the .eslinrc file in the repo.
choose whichever works for you :D (and have the highest number or downloads)

frontend/src/app/components/header.js
Fezzzi: Done and done

frontend/src/app/components/theme.js
Fezzzi: @petrk-salsita exmplained to me, I would create another mini PR to add lint into scripts in `package.json` file and to resolve any issues that might pop out form it.

frontend/src/app/modules/root/root-reducer.js
```diff
@@ -14,13 +15,13 @@ const initialState = {
   users: [],
   activeAlert: null,
   activeRedirect: null,
-  theme: window.localStorage.getItem("theme") || "dark",
+  theme: window.localStorage.getItem('theme') || 'dark',
```
petrk-salsita: `ThemeTypes.Dark` can be used instead of the string. Also maybe extract the `'theme'` item name into constant as it's used multiple times?

frontend/src/app/modules/theme/theme-effects.js
```diff
@@ -0,0 +1,11 @@
+import { ThemeTypes } from '../../const/theme-types'
+
+export const changeTheme = theme => {
+  const newTheme = theme === ThemeTypes.Dark ? ThemeTypes.Light : ThemeTypes.Dark
+  window.localStorage.setItem('theme', newTheme)
+
+  return {
+    theme: newTheme,
+    themeTransition: 1,
```
petrk-salsita: The transition is reset to `1` upon every theme change - it probably doesn't need to be part of the payload, the reducer can just set it automatically as a reaction to the `themeChanged` action. Additionally, it might be better to use the `boolean` values `true` & `false` for this purpose.

Fezzzi: I agree, will do

frontend/src/app/modules/root/root-reducer.js
Fezzzi: I disagree with the necessity for the `'theme'` string constant. There are 2 usages of the string and I think it would decrease readability more than it would increase reusability. Not to mention it would require to name the constant which is pretty time consuming :( What do you think @petrk-salsita ?

frontend/src/app/modules/theme/theme-effects.js
Fezzzi: [done](https://github.com/salsita/foosball-rating/pull/81/commits/6233538904b5337673c199255c73c62943d1415e#diff-a40f089e71d55fac58c940a9aa91b1d0)

frontend/src/app/modules/root/root-reducer.js
Fezzzi: [done](https://github.com/salsita/foosball-rating/pull/81/commits/6233538904b5337673c199255c73c62943d1415e#diff-a40f089e71d55fac58c940a9aa91b1d0)

frontend/src/app/modules/root/root-reducer.js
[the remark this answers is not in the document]
Fezzzi: [done](https://github.com/salsita/foosball-rating/pull/81/commits/25ba48df7bd72b6d383bdf8f576df28eb479576d#diff-33a3300b9cf6853c180e26ffe5828ca6)

frontend/src/app/components/header.js
Fezzzi: Resolved [here](https://github.com/salsita/foosball-rating/pull/81/commits/25ba48df7bd72b6d383bdf8f576df28eb479576d#diff-bf7b3488ddec0115f2387242335ffae0)

> you also don't need to write _window_ explicitly

done [here](https://github.com/salsita/foosball-rating/pull/81/commits/25ba48df7bd72b6d383bdf8f576df28eb479576d#diff-a5b8f7d4dfd7d1c2c65b70a166a77b8f) and [here](https://github.com/salsita/foosball-rating/pull/81/commits/25ba48df7bd72b6d383bdf8f576df28eb479576d#diff-bf7b3488ddec0115f2387242335ffae0) , although I still use window during acessing localStorage. Shall I remove this as well @novellizator?

frontend/src/app/components/theme.js
Fezzzi: [done](https://github.com/salsita/foosball-rating/pull/81/commits/93d14d9d4dfbc131f4c797ffe52478fa61f2614f) and [created](#90 )

frontend/src/app/components/theme.js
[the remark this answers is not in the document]
Fezzzi: done [here](https://github.com/salsita/foosball-rating/pull/81/commits/25ba48df7bd72b6d383bdf8f576df28eb479576d#diff-7c47b25cde5324f01910a9b3a770075e)

frontend/src/app/modules/theme/theme-actions.js
```diff
@@ -1,7 +1,7 @@
 import { createActions } from 'reduxsauce'
 
 export const ThemeActions = createActions({
-  themeChanged: ['newTheme'],
+  themeChanged: ['theme'],
   changeTheme: ['theme'],
```
petrk-salsita: Maybe change the argument name to 'currentTheme' or something similar - just by seeing this, I'd think that the 'theme' argument holds the 'theme' you want to change to.

Fezzzi: done [here](https://github.com/salsita/foosball-rating/pull/81/commits/44f0448e166dac047262d9a907d4cc973e9a7619#diff-ebe66918bbf9967140fef8b0899e13f1)

frontend/src/app/modules/root/root-reducer.js
Fezzzi: added global constants [here](https://github.com/salsita/foosball-rating/pull/81/commits/44f0448e166dac047262d9a907d4cc973e9a7619#diff-fc024cf3f843472eca76dc473839b5c8)

PR #89, opened 2020-02-14
Add marks to the plot line to make it clear when a match was played

zbrcz:
Recreation of https://github.com/salsita/foosball-rating/pull/70.


zbrcz:
With 50+ matches, the marks become obtrusive so this can't be merged as it is.

It could be solved by displaying just one or two months of the graph and making it horizontally scrollable.

novellizator:
Closed because of https://github.com/salsita/foosball-rating/pull/89#issuecomment-586312112

PR #95, opened 2020-02-27
Dotenv config for frontend

Fezzzi:
Testing the docker edit via Deployment...


Fezzzi:
Seems it didn't broke anything, can you CR @novellizator @petrk-salsita please?

frontend/.env.local.example
```diff
@@ -0,0 +1,2 @@
+REACT_APP_POST_URL_PORT=":3000"
```
novellizator: i'd leave out the colon here

README.md
```diff
@@ -7,17 +7,7 @@ Notes on running the app locally
...
-    +++ b/frontend/src/app/modules/api/client-config.js
-    @@ -1,4 +1,4 @@
-     export const config = {
-    -    baseURL: `${window.location.protocol}//${window.location.hostname}/api`,
-    +    baseURL: `${window.location.protocol}//${window.location.hostname}:3000`,
-         timeout: 5000
-     }
-    ```
+2) Copy `frontend/.env.local.example` to `frontend/.env` 
+(does not apply in case of runnig in docker)
```
novellizator: running

Fezzzi: Done [here](https://github.com/salsita/foosball-rating/pull/95/commits/0a92aa2b8d648b6df6e9301cec83602f84b58d28#diff-04c6e90faac2675aa89e2176d2eec7d8), sorry about that

frontend/.env.local.example
Fezzzi: Fixed [here](https://github.com/salsita/foosball-rating/pull/95/commits/0a92aa2b8d648b6df6e9301cec83602f84b58d28#diff-5696aa6f6ae27f14a99322ab211068ba)

frontend/src/app/modules/api/client-config.js
```diff
@@ -1,5 +1,7 @@
 export const config = {
   baseURL: `${window.location.protocol}//${window.location.hostname}` +
-    `${process.env.REACT_APP_POST_URL_PORT}${process.env.REACT_APP_POST_URL_PATH_BASE}`,
+    `${process.env.REACT_APP_POST_URL_PORT !== ""
```
novellizator: I guess, the whole point of having backticks is so that you don't use "+" for concatenation :D 
Also: 1. consider making an introductory variable 

```
const envPort: string = process.env.REACT_APP_POST_URL_PORT || '';
const colonPortOrNothing:string = envPort ? `:${envPort}`: ''
```
Or something similar :) 
Also, you are testing for a string equality against an empty string. This is quite fragile. What if the REACT_APP_POST_URL_PORT is not set, hence undefined?

frontend/.env.local.example
```diff
@@ -1,2 +1,2 @@
-REACT_APP_POST_URL_PORT=":3000"
+REACT_APP_POST_URL_PORT="3000"
```
novellizator: pls add a new line, even github suggests you that - to soothe my OCD

Fezzzi: [here](https://github.com/salsita/foosball-rating/commit/eca8143fe7504de0175d2608f3b8d0f0f03bd9e7#diff-8c572d46d23eaf9d608eec00f24c16a3)

frontend/src/app/modules/api/client-config.js
Fezzzi: Great idea, done [here](https://github.com/salsita/foosball-rating/commit/eca8143fe7504de0175d2608f3b8d0f0f03bd9e7#diff-5696aa6f6ae27f14a99322ab211068ba)

I am using backticks mostly for their clear interpolation. The contatenation prevented whitespaces from breaking the URL. But I agree this way it looked rather ugly.

PR #97, opened 2020-03-05
Rename backend js to ts files

### LINKED ISSUE #96: Convert backend js to ts files
Only renaming and typescript setup. Code should be left in the same state. Part of Typescript migration. If we make more changes then git rebase or merge won't notice conflicts. When all files are ts already conflicts will be visible better (auto-merging still can break things without notice - hence unit tests exist ;))

littlewhywhat:
Closes #96

Only wires Typescript and renames files. Checks are ignored via --log-error option


backend/app.ts
```diff
@@ -1,13 +1,13 @@
 const express = require('express')
 const bodyParser = require('body-parser')
 
-const storage = require('./storage/storage')
-const matchRepository = require('./repositories/match-repository')
-const userRepository = require('./repositories/user-repository')
+const storage = require('./storage/Storage')
```
novellizator: why the rename?

littlewhywhat: It seems like this convention is used in https://github.com/salsita/performance-review. It was also used in the PR that I am breaking up. Do we have some common convention defined somewhere? Should I keep file names as they were before?

petrk-salsita: I don't have a strong preference 🙂 Feel free to keep the files renamed as long as it makes sense to you (if @novellizator is ok with that of course).

novellizator: I'm fine with this

PR #100, opened 2020-03-09
gh-98: Remove --log-error, fix imports

### LINKED ISSUE #98: Solve typescript errors on the backend
Solving only errors reported by ts-node to remove --log-error flag

littlewhywhat:
Closes #98 

Also had to declare properties

### DIFF
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -1,13 +1,13 @@
-const express = require('express')
-const bodyParser = require('body-parser')
+import * as express from 'express'
+import * as bodyParser from 'body-parser'
 
-const storage = require('./storage/Storage')
-const matchRepository = require('./repositories/MatchRepository')
-const userRepository = require('./repositories/UserRepository')
+import * as storage from './storage/Storage'
+import * as matchRepository from './repositories/MatchRepository'
+import * as userRepository from './repositories/UserRepository'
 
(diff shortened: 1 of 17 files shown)


backend/app.ts
```diff
@@ -1,13 +1,13 @@
-const express = require('express')
-const bodyParser = require('body-parser')
+import * as express from 'express'
```
petrk-salsita: Just a theoretical question - we might want to replace this by importing just the objects that are being used 🤔 But that can of course be done in the future in a separate PR.

littlewhywhat: I think for our modules it will be done in future PR. Regarding 3rd party - actually I realize now that it should be possible to import, for example, "express" without *. However I struggle with this and get errors about default import or that express is not a function :(

littlewhywhat: I see from performance-review project that probably you meant smth like this:
```
import {
  Application, NextFunction, Request, Response,
} from 'express'

const express: Application = express()
```
Yes, I think it can be done other PR when we get lint rules for reporting it as a warning or error.

littlewhywhat: it seems ok to import express with * or use require - Microsoft does it here with * https://github.com/microsoft/TypeScriptSamples/blob/master/imageboard/app.ts. So I guess I can merge then :)

PR #101, opened 2020-03-09
gh-67: Add eslint to backend

### LINKED ISSUE #67: As a programmer I want a linter on BE
now it's embarrasing how many turns the reviews are taking just to fix the " vs ' inconsistencies.

littlewhywhat:
Closes #67

Extracted common eslint config to the root.
Grouped devDependencies

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,59 @@
+{
+  "root": true,
+  "env": {
+    "node": true
+  },
+  "rules": {
+    "no-var": "error",
+    "prefer-arrow-callback": "warn",
+    "no-unused-vars": [
+      "warn"
+    ],
+    "no-warning-comments": "warn",
+    "no-console": 0,
+    "max-len": [
+      "warn",
+      {
+        "code": 100,
+        "ignoreComments": true,
+        "ignoreUrls": true,
+        "ignoreStrings": true,
+        "ignoreTemplateLiterals": true,
+        "ignoreRegExpLiterals": true
+      }
+    ],
+    "comma-dangle": [
+      "error",
+      "always-multiline"
+    ],
+    "eol-last": [
+      "error",
+      "always"
+    ],
+    "keyword-spacing": [
+      "error"
+    ],
+    "object-curly-spacing": [
+      "error",
+      "always"
+    ],
+    "no-trailing-spaces": "error",
+    "prefer-const": "warn",
+    "operator-linebreak": [
+      "error",
+      "after",
+      {
+        "overrides": {
+          "?": "before",
+          ":": "before",
+          "===": "ignore"
+        }
+      }
+    ],
+    "arrow-parens": [
+      "warn",
+      "as-needed"
+    ],
+    "consistent-return": "error"
+  }
+}
--- a/backend/.eslintrc.json
+++ b/backend/.eslintrc.json
@@ -1,0 +1,29 @@
+{
+  "parser": "@typescript-eslint/parser",
+  "plugins": [
+    "@typescript-eslint"
+  ],
+  "extends": [
+    "plugin:@typescript-eslint/recommended"
+  ],
+  "rules": {
+    "quotes": "off",
+    "@typescript-eslint/quotes": [
+      "error",
+      "single"
+    ],
+    "semi": "off",
+    "@typescript-eslint/semi": [
+      "error",
+      "never"
+    ],
+    "indent": "off",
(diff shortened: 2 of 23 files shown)


.eslintrc.json
```diff
@@ -0,0 +1,58 @@
+{
```
novellizator: i'd add ` "root": true` here, see https://eslint.org/docs/user-guide/configuring

.eslintrc.json
```diff
@@ -0,0 +1,58 @@
...
+        }
+      }
+    ],
+    "arrow-parens": [
+      "warn",
+      "as-needed"
+    ],
+    "consistent-return": "error"
+  }
+}
```
novellizator: i'd add a trailing new line (and configure it in your IDE). -> you can see the red arrow pointing at that...

frontend/.eslintrc.json
```diff
@@ -29,69 +28,20 @@
         "never"
       ],
       "mocha/no-exclusive-tests": "error",
-      "no-var": "error",
-      "prefer-arrow-callback": "warn",
-      "no-unused-vars": "warn",
-      "no-warning-comments": "warn",
-      "no-console": 0,
       "indent": [
```
novellizator: why not keeping the indent rule in the root?

backend/.eslintrc.json
```diff
@@ -0,0 +1,29 @@
...
+    "@typescript-eslint/quotes": [
+      "error",
+      "single"
+    ],
+    "semi": "off",
+    "@typescript-eslint/semi": [
+      "error",
+      "never"
+    ],
+    "indent": "off",
```
novellizator: indent off?

littlewhywhat: see docs https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/indent.md#options. To avoid incorrect errors I suppose

frontend/.eslintrc.json
littlewhywhat: it seems like those are different rules - you can't use the eslint rule for typescript. I actually did put all in root but then I noticed that it didn't report me indent for class properties/methods correctly. Therefore for each eslint basic rule I decided to prefer a same named rule in typescript plugin if it exists.

.eslintrc.json
littlewhywhat: do you mean eol-last https://eslint.org/docs/rules/eol-last? It's already there

novellizator: if you notice this file in the code review, you see the marker, that you're missing this in this very file ;)

.eslintrc.json
littlewhywhat: fixed

.eslintrc.json
littlewhywhat: fixed

littlewhywhat:
> Also - lifting some of the rules to the parent dir - doesn't that require some linter fixes? (did you run eslint --fix ?)
I ran on both backend and frontend - there are some auto fixes to be made on both actually - frontend is easy to resolve though but it's preexisting. I would autofix everything in a separate PR if there are no objections

littlewhywhat:
Fixing MatchReporter I have some errors on the rule "no-use-before-declare". It's actually discouraged in the modern Typescript - see https://palantir.github.io/tslint/rules/no-use-before-declare. @novellizator do we want to keep this one? To fix its errors I would need to reorder the functions in the MatchReporter in future.

littlewhywhat:
> Fixing MatchReporter I have some errors on the rule "no-use-before-declare". It's actually discouraged in the modern Typescript by palantir - see https://palantir.github.io/tslint/rules/no-use-before-declare. Yet it is included to recommended config. @novellizator @petrk-salsita do we want to keep this one? To fix its errors I would need to reorder the functions in the MatchReporter in future.

Turns out it 'no-use-before-define' and we will be keeping it for now. Thanks!

littlewhywhat:
> > Also - lifting some of the rules to the parent dir - doesn't that require some linter fixes? (did you run eslint --fix ?)
> 
> I ran on both backend and frontend - there are some fixes to be made on both actually - frontend could be autofixed and it's preexisting - maybe we should add an issue to include eslint in build pipeline to avoid necessity to double check. I would autofix everything in a separate PR if there are no objections

Ok I add autofix. Rules are fixed for this PR now

littlewhywhat:
> well done. please scope together with autofixes

added autofixes

Issue #109, opened 2020-03-11
Use node-pg-migrate instead of current tools


littlewhywhat:
starting to use the node-pg-migrate instead of what's currently there

https://github.com/salsita/node-pg-migrate
take a look at how it's used in performance review app -  https://github.com/salsita/performance-review/pull/37/files

part of https://github.com/salsita/foosball-rating/issues/107

PR #110, opened 2020-03-11
gh-109: Use node-pg-migrate

littlewhywhat:
Closes gh-109


littlewhywhat:
I didn't convert test_data.sql to ts because I think it's easier to maintain in sql format then as string in ts. It happened that node-pg-migrate works well with sql as well. Let me know what you think.

Issue #107, opened 2020-03-11
Change the DB model and add migration


littlewhywhat:
Add GameTypes table and add a foreign key column for Matches table.

Add migrations to create Foosball game type and update existing Matches records with a Foosball game type id.

Hardcode Foosball game type id in insert query for new Match

PR #113, opened 2020-03-13
gh-112: Add API for games

### LINKED ISSUE #112: Add api to add new game type

littlewhywhat:
Closes #112 
After #107

Besides following the trends of the existing code also added some unit tests for new logic.

### DIFF
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -1,9 +1,11 @@
 import * as express from 'express'
+import { Response, Request } from 'express'
 import * as bodyParser from 'body-parser'
 
 import * as storage from './storage/Storage'
 import * as matchRepository from './repositories/MatchRepository'
 import * as userRepository from './repositories/UserRepository'
+import { addGame } from './repositories/GameRepository'
 
 import { makeBot, SingleChannelBot } from './bot/bot-factory'
 
@@ -53,6 +55,12 @@ app.get('/matches', (req, res) => {
 
 })
 
+app.get('/games', (req: Request, res: Response) => {
+  storage.getAllGames()
+    .then(res.send.bind(res))
+    .catch(error => processError(res, error))
+})
+
 app.post('/users', (req, res) => {
   userRepository.addUser(req.body)
     .then(res.send.bind(res))
@@ -74,6 +82,12 @@ app.post('/matches', async (req, res) => {
   } catch (error) {
     processError(res, error)
   }
+})
+
+app.post('/games', (req: Request, res: Response) => {
+  addGame(req.body)
+    .then(res.send.bind(res))
+    .catch(error => processError(res, error))
 })
 
 app.listen(port, () => console.log(`Foosball backend running on ${port}!`))
--- a/backend/repositories/GameRepository.ts
+++ b/backend/repositories/GameRepository.ts
@@ -1,0 +1,19 @@
+import { GameData } from '../types/Game'
+import { InputError } from '../errors/InputError'
+import * as storage from '../storage/Storage'
+
+const isValidGame = (gameData: GameData): boolean => {
+  return gameData.name &&
+    gameData.description &&
+    gameData.name.trim() == gameData.name &&
+    gameData.name.toLowerCase() == gameData.name &&
+    !gameData.name.includes(' ') &&
+    gameData.description.trim() == gameData.description
+}
+
+export const addGame = async (gameData: GameData): Promise<void> => {
+  if (!isValidGame(gameData)) {
+    throw new InputError('Game data is not valid!')
+  }
+  storage.insertGame(gameData)
+}
--- a/backend/repositories/GameRepository.ts
+++ b/backend/repositories/GameRepository.ts
@@ -1,19 +1,22 @@
-import { GameData } from '../types/Game'
+import { GameData, isGameData } from '../types/Game'
 import { InputError } from '../errors/InputError'
 import * as storage from '../storage/Storage'
 
-const isValidGame = (gameData: GameData): boolean => {
-  return gameData.name &&
-    gameData.description &&
+const isValidGameData = (gameData: GameData): boolean => {
+  return gameData.name !== '' &&
+    gameData.description !== '' &&
     gameData.name.trim() == gameData.name &&
     gameData.name.toLowerCase() == gameData.name &&
     !gameData.name.includes(' ') &&
     gameData.description.trim() == gameData.description
 }
 
-export const addGame = async (gameData: GameData): Promise<void> => {
-  if (!isValidGame(gameData)) {
+export const addGame = async (data: unknown): Promise<void> => {
+  if (!isGameData(data)) {
+    throw new InputError('Data is not valid!')
+  }
+  if (!isValidGameData(data)) {
     throw new InputError('Game data is not valid!')
   }
-  storage.insertGame(gameData)
+  storage.insertGame(data)
 }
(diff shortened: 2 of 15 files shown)


backend/tests/StorageContextTest.ts
```diff
@@ -0,0 +1,69 @@
+import { StorageContext } from '../storage/StorageContext'
```
novellizator: convention thing: when we write unit tests, we usually have the tests in the same directory as normal files, only with a different (".test.ts") suffix. so I you have /backend/blah/StorageContext.ts, the tests reside in /backend/blah/StorageContext.test.ts
I honestly like that better, since you usually edit tests along with the tested file and they are quite tighly coupled.

backend/app.ts
```diff
@@ -12,7 +14,7 @@ import { MatchReporter } from './match-reporter/MatchReporter'
 const jsonParser = bodyParser.json()
 const urlencodedParser = bodyParser.urlencoded({ extended: false })
 
-const app = express()
+const app: Application = express()
```
novellizator: i'd expect timescript to be ablt to infer the time of express() - isn't that the case? 🤔

backend/app.ts
```diff
@@ -12,7 +14,7 @@ import { MatchReporter } from './match-reporter/MatchReporter'
...
 const port = 3000
```
novellizator: just as as aside - it would be good to stick it into an env in the future

backend/app.ts
littlewhywhat: you are right

backend/tests/StorageContextTest.ts
littlewhywhat: I am more used to java like style where tests end up in a separate folder. This way it actually reminds me more C++ style! :) ok, I guess I don't mind either approach... and where do we store some test utils modules in this case? I tend recently to put some common util functions and data for tests into common modules and then import them in each test.

backend/repositories/GameRepository.ts
```diff
@@ -0,0 +1,25 @@
+import { GameData } from '../types/Game'
+import { InputError } from '../errors/InputError'
+import * as storage from '../storage/Storage'
+
+const isValidGame = (gameData: GameData): boolean => {
+  if (!gameData.name || !gameData.description) {
```
novellizator:
```
if (X) {
  return false
} else if (Y) {
  return false
}
```

could be further improved as 
```
if (X || Y) {
  return false
}
```

backend/repositories/GameRepository.ts
```diff
@@ -0,0 +1,25 @@
...
+  } else if (gameData.name.includes(' ')) {
+    return false
+  } else if (gameData.description.trim() != gameData.description) {
+    return false
+  }
+  return true
+}
+
+export const addGame = async (gameData: GameData): Promise<void> => {
+  if (!isValidGame(gameData)) {
```
novellizator: why are you allowing a creation of an invalid game? why not validate it right at the time of construction? (ie https://github.com/salsita/foosball-rating/pull/113/files#diff-382e120a6484790f7829e0ec3bb18713R9)

backend/storage/StorageContext.ts
```diff
@@ -129,6 +130,34 @@ export class StorageContext {
...
+        game.description,
+      ])
+    } catch (error) {
+      console.error(error)
+      if (dbErrors.isUniqueViolation(error)) {
+        throw new ConflictError('Game already exists')
+      }
+      throw new Error('Unable to insert new game')
+    }
+    return row !== null ? dbTransformations.createGameFromDbRow(row): null
```
novellizator: `x !== null ? FOO : null` could we simplified into `x && FOO`

backend/tests/StorageContextTest.ts
novellizator: maybe those should go to the aforementioned /test folder :)

backend/repositories/GameRepository.ts
littlewhywhat: GameData is just description of the object created from body of the request - it's created by express.

backend/tests/StorageContextTest.ts
littlewhywhat: fixed in https://github.com/salsita/foosball-rating/pull/113/commits/d8bb46a15e9ce4e8733aba65b651d0ec0eece570

backend/repositories/GameRepository.ts
littlewhywhat: fixed in https://github.com/salsita/foosball-rating/pull/113/commits/d8bb46a15e9ce4e8733aba65b651d0ec0eece570

backend/storage/StorageContext.ts
littlewhywhat: fixed in https://github.com/salsita/foosball-rating/pull/113/commits/d8bb46a15e9ce4e8733aba65b651d0ec0eece570

backend/app.ts
littlewhywhat: filed https://github.com/salsita/foosball-rating/issues/127

backend/repositories/GameRepository.ts
novellizator: hmm then this is an antipattern. you're supplying any into addGame - you don't know if the argument has a given type.

here's what I'd suggest:

```

interface GameData {
}

// notice the test IS GameData as a return postcondition
function isGameData(test: unknown): test is GameData {
    // use the conditions from your isValidGame
    // if this function returns true is means that the tested variable has the GameDataType
    return "the conditions from your in isValidGame" != test;
}

function addGame(maybeGameData: unknown) {
  if (!isGameData(maybeGameData)) {
    throw Error("Not game info!")
  }

  // insert into storage
}



```

backend/types/Game.ts
```diff
@@ -0,0 +1,14 @@
+export class Game {
```
novellizator: you don't instantiate either --> either use it, or simply turn it into an interface

backend/repositories/GameRepository.ts
novellizator: more can be found here https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

novellizator: also, it's weird that the TS compiler didn't protest against you supplying any. maybe i'd tighten the screws, so to speak (you can have it as a separate issue in the future - https://stackoverflow.com/questions/32227277/typescript-typed-function-argument-accepts-any )

backend/types/Game.ts
littlewhywhat: Fixed in https://github.com/salsita/foosball-rating/pull/113/commits/f657e38aac9173ce08efb3ec3cf4c740fc7b689b

backend/repositories/GameRepository.ts
littlewhywhat: antipattern fixed in https://github.com/salsita/foosball-rating/pull/113/commits/f657e38aac9173ce08efb3ec3cf4c740fc7b689b

littlewhywhat: issue filed for 'any' https://github.com/salsita/foosball-rating/issues/133

littlewhywhat:
Merged validation methods into one https://github.com/salsita/foosball-rating/pull/113/commits/a2f588b635e29cf438b77cbf2aded03a3c25fb59

PR #114, opened 2020-03-13
Issue 27 enhance statistics for user

### LINKED ISSUE #27: Enhance the statistics for a user
Suggested features:
- most frequent team members
- least frequent team members
- same for opponents
- most likely to win with / lose with
etc...

Fezzzi:
I decided for a simple table design under the graph. 

Both design and functionality remarks are welcome!

Closes #27


frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +77,104 @@ export const computeWinRatio = (playerId, playerMatches) => {
   const wonMatchesCount = playerMatches.filter(match => didPlayerWin(playerId, match)).length
   return (playerMatches.length > 0) ? (wonMatchesCount / playerMatches.length) : 0
 }
+
+export const computeDays = matchChanges => {
+  let worstDayKey = ''
+  let bestDayKey = ''
+  const daysMap = {}
+  for (const match of matchChanges) {
```
littlewhywhat: personally I would do it in two passes for readability - first create a map ratingChangesByDate or smth like that and then compute the max/min of the values. Or maybe I would vote to put a little note how the computation goes in one pass.

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +77,104 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+  const opponents = findSignificanPlayers(playerId, playerMatches, getOpponents)
+  return {
+    mostFrequentOpponent: opponents.mostFrequentPlayer,
+    leastFrequentOpponent: opponents.leastFrequentPlayer,
+    mostSuccessOpponent: opponents.mostSuccessPlayer,
+    leastSuccessOpponent: opponents.leastSuccessPlayer,
+  }
+}
+
+const findSignificanPlayers = (playerId, playerMatches, playersProvider) => {
```
littlewhywhat: typo in the name

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +77,104 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+  const playersMap = {}
+  for (const match of playerMatches) {
+    const players = playersProvider(playerId, match)
+    for (const player of players) {
+      if (!Object.hasOwnProperty.call(playersMap, player.id)) {
+        playersMap[player.id] = {
+          player: player,
+          matches: 1,
+          wins: Number(didPlayerWin(playerId, match)),
+          winRatio: Number(didPlayerWin(playerId, match)),
```
littlewhywhat: Is this winRatio used somehow? it doesn't seem to be recomputed...

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +77,104 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+  }
+}
+
+const findSignificanPlayers = (playerId, playerMatches, playersProvider) => {
+  let mostFrequentPlayer = {}
+  let leastFrequentPlayer = {}
+  let mostSuccessPlayer = {}
+  let leastSuccessPlayer = {}
+  const playersMap = {}
+  for (const match of playerMatches) {
```
littlewhywhat: as for computeDays - I would separate these in several passes for readability and scaling - one day we may need to have just a frequent player and we would need either to write it from scratch or somehow extract it from here with options to add bugs :(

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +77,104 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+      }
+
+      mostFrequentPlayer = updateMaxFromMap(mostFrequentPlayer, 'matches', playersMap[player.id])
+      leastFrequentPlayer = updateMinFromMap(leastFrequentPlayer, 'matches', playersMap[player.id])
+      mostSuccessPlayer = updateMaxFromMap(mostSuccessPlayer, 'wins', playersMap[player.id])
+      leastSuccessPlayer = updateMaxFromMap(leastSuccessPlayer, 'losses', playersMap[player.id])
+    }
+  }
+
+  return {
```
littlewhywhat: what if there are not wins or no loses or no matches? is it ok to return empty objects? I am afraid links can break

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +77,104 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+    } else {
+      daysMap[key]['value'] += value
+    }
+
+    bestDayKey = updateMaxFromMap(daysMap[bestDayKey] ?? {}, 'value', daysMap[key]).date
+    worstDayKey = updateMinFromMap(daysMap[worstDayKey] ?? {}, 'value', daysMap[key]).date
+  }
+
+  return {
+    bestDay: daysMap[bestDayKey],
```
littlewhywhat: we may return empty object if there are no matches, is it expected?

frontend/src/app/modules/matches/matches-computations.js
littlewhywhat: also same as with computeDays - maybe precompute data and then just run min, max algorithm - https://stackoverflow.com/questions/4020796/finding-the-max-value-of-an-attribute-in-an-array-of-objects - on a prepared array if possible

littlewhywhat: preprocessing could also let you to avoid using playersProvider callback - you could have something like an array of all opponents and array of all teammates and then do computations on them. Again I think it would be more extensible in future.

frontend/src/app/components/ProfileStatistics/index.js
```diff
@@ -0,0 +1,76 @@
...
+            <td>Best Day:
+              <StatisticsValue>{bestDay?.date} {getDayScorePart(bestDay?.value)}</StatisticsValue>
+            </td>
+            <td>Worst Day:
+              <StatisticsValue>{worstDay?.date} {getDayScorePart(worstDay?.value)}</StatisticsValue>
+            </td>
+          </tr>
+          <tr>
+            {getPlayerRow('Most Played With', mostFrequentTeammate, 'matches', true)}
+            {getPlayerRow('Least PLayed With', leastFrequentTeammate, 'matches', false)}
```
littlewhywhat: typo

Fezzzi:
> Sorry quite a lot of comments - feature seems to be simple but it's actually not :) in general the logic for computations is quite complicated and we should be careful with passing props to selectors in my opinion. Anyway I think it's a good start! Let me know what you think about my comments

Understandable amount of comments regarding the code quality I delivered at first. Sorry about that, I did it in a hurry between working on other project. Uploading newever version.

Link to player profile still don't work, will wix in next commit by wrapping `createProfilePath` with `constructUrl` in `StatsRow`.

frontend/src/app/components/ProfileStatistics/index.js
Fezzzi: Fixed [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-744525510377a0aca19d17cf4dc35116)

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Fixed by providing default values [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-304dca28d92d292252e86b7e83fc1569)

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Fixed by providing default values [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-304dca28d92d292252e86b7e83fc1569)

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Did the thing with the min/max approach [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-304dca28d92d292252e86b7e83fc1569) and [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-08117b849ecc00af7c3c0d609dd86d31).

I don't like the idea of running through the matches and seperating players into structures of teammates and opponents and then running through them for actual computing. It would IMHO be less readable than a simple function that extracts either opponents or teammates but I am open to discussion in this...

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Nope, have no idea how it got there, removed [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-304dca28d92d292252e86b7e83fc1569)

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Completely renamed [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-304dca28d92d292252e86b7e83fc1569)

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Done [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-08117b849ecc00af7c3c0d609dd86d31)

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +77,104 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+const findSignificanPlayers = (playerId, playerMatches, playersProvider) => {
+  let mostFrequentPlayer = {}
+  let leastFrequentPlayer = {}
+  let mostSuccessPlayer = {}
+  let leastSuccessPlayer = {}
+  const playersMap = {}
+  for (const match of playerMatches) {
+    const players = playersProvider(playerId, match)
+    for (const player of players) {
+      if (!Object.hasOwnProperty.call(playersMap, player.id)) {
```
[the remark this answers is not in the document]
Fezzzi: Personally, I don't like maps due to the necessity to always call `get` and `set` instead of direct access. Also, it would require some more edits as the function will now return map instead of object. I made the code simpler with the second approach you mentioned [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-304dca28d92d292252e86b7e83fc1569)

frontend/src/app/pages/Profile/index.js
```diff
@@ -27,10 +28,12 @@ class ProfileComponent extends Component {
         <Subtitle>Elo rating: {rating}</Subtitle>
         <ProfileDetail>
           <Subtitle>Matches: {totalMatches}</Subtitle>
+          <Subtitle>Wins: {Math.round(totalMatches * winRatio)}</Subtitle>
```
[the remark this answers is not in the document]
Fezzzi: I guess it would have to be like hundred or thousands of matches for the rounding error to possibly appear but I agrre it is a bad practise so I changed it to [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-08117b849ecc00af7c3c0d609dd86d31)

frontend/src/app/modules/players/players-selectors.js
```diff
@@ -62,3 +62,18 @@ const getPlayerCriteriaPoints = (player, matchesLast, criteria) => {
...
+  getTopPlayers,
+  (state, playerId) => playerId,
+  (players, playerId) => getRatingStatisticsForPlayer(players, playerId),
+)
+
+const getRatingStatisticsForPlayer = (players, playerId) => {
+  const index = players.findIndex(player => player.id === playerId)
+  return {
+    ranking: index + 1,
+    toNextRank: index > 0 ? players[index - 1].rating - players[index].rating : 0,
```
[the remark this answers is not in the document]
Fezzzi: rank is position detoned by the `getTopRatedPlayers` which is also used in players listing (surrounded by all thhis stuff from filters). Therefore, regardless of the rating there is always someone that will be listed first, second, etc...

frontend/src/app/modules/players/players-selectors.js
```diff
@@ -62,3 +62,18 @@ const getPlayerCriteriaPoints = (player, matchesLast, criteria) => {
       return player.rating
   }
 }
+
+export const getRankingsForPlayer = createSelector(
+  getTopPlayers,
+  (state, playerId) => playerId,
+  (players, playerId) => getRatingStatisticsForPlayer(players, playerId),
```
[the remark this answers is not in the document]
Fezzzi: Unfortunatelly not, as there would be some problem with calling `.apply` during memoization that I don't really understand. I think it is because `getRatingStatisticsForPlayer` is a function and not a selector

frontend/src/app/modules/players/players-selectors.js
```diff
@@ -62,3 +62,18 @@ const getPlayerCriteriaPoints = (player, matchesLast, criteria) => {
       return player.rating
   }
 }
+
+export const getRankingsForPlayer = createSelector(
+  getTopPlayers,
```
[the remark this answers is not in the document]
Fezzzi: Changed to `getTopRatedPlayers` [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-cd4220e6543e26c82876c33466c868e0)

I don't really understand the memoization to the degree that will allow me to have any opinion about this so if you propose any change to this particular piece of code i will do it.

frontend/src/app/modules/players/players-selectors.js
```diff
@@ -62,3 +62,18 @@ const getPlayerCriteriaPoints = (player, matchesLast, criteria) => {
       return player.rating
   }
 }
+
+export const getRankingsForPlayer = createSelector(
```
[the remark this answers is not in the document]
Fezzzi: The same as in the message bellow...

frontend/src/app/modules/matches/matches-selectors.js
```diff
@@ -37,11 +40,21 @@ const getLastMatchesForPlayer = createSelector(
...
+  }
+}
+
+const computeMoreStatistics = (playerId, matchChanges, playerMatches) => ({
+  ...computeDays(matchChanges),
+  ...computeTeammates(playerId, playerMatches),
+  ...computeOpponents(playerId, playerMatches),
 })
 
 export const getStatisticsForPlayer = createSelector(
```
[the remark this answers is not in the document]
Fezzzi: Changed this in [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-08117b849ecc00af7c3c0d609dd86d31), that should memoize fine, right?

frontend/src/app/pages/Profile/index.js
```diff
@@ -27,10 +28,12 @@ class ProfileComponent extends Component {
         <Subtitle>Elo rating: {rating}</Subtitle>
         <ProfileDetail>
```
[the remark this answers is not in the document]
Fezzzi: done [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-08117b849ecc00af7c3c0d609dd86d31)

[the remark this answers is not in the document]
Fezzzi: Agree, I moved stats up [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-07a1f0d168a8c0564c021e065282760a) as well

frontend/src/app/components/ProfileRatingGraph/index.js
```diff
@@ -22,7 +22,7 @@ const SimpleProfileRatingGraph = props => (
       <FlexibleWidthXYPlot
         height={300}
         xType='time'
-        margin={{ left: 45 }}
+        margin={{ left: 45, bottom: 65 }}
```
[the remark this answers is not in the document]
Fezzzi: This was done because by the time of former edits, month names weren't truncated and `September` and `November` weren't printed corretly. I removed that [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-7f8fb954a84804e36f1c0e7e6293ddf3) as I see month names are truncated now.

frontend/src/app/components/ProfileStatistics/index.js
```diff
@@ -0,0 +1,76 @@
...
+            {getPlayerRow('Most Won Over', mostSuccessOpponent, 'wins', true)}
+            {getPlayerRow('Most Lost To', leastSuccessOpponent, 'losses', false)}
+          </tr>
+        </tbody>
+      </StatisticsTable>
+    </>
+  )
+}
+
+const getDayScorePart = score => getScorePart(score, score > 0, score > 0 ? '+' : '')
```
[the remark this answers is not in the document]
Fezzzi: Added new component for taht [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-9408a2f2c965bbede4015953e5ee877c)

frontend/src/app/components/ProfileStatistics/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+const getScorePart = (value, isGreen, sign = 'x') => (
+  <WinnerSpan winner={isGreen}>{sign}{value}</WinnerSpan>
```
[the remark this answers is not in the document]
Fezzzi: Added new kind of span - StatsSpan [here](https://github.com/salsita/foosball-rating/pull/114/commits/3d398530cdf75f980e6344e885e4f18dd919311a#diff-30db3d054850af4aa6f2ec06875af6ba)

frontend/src/app/components/ProfileStatistics/index.js
```diff
@@ -0,0 +1,76 @@
...
+    },
+  } = props
+
+  return (
+    <>
+      <TextSpan>Ranking: {ranking}</TextSpan>
+      <StatisticsTable>
+        <tbody>
+          <tr>
+            <td>To Next Rank:
```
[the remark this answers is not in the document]
Fezzzi: I prefer former solution as the layout of each cell is very simple, it contains only text aligned to left and text aligned to right but I guess I could wrap each of them in some `TextSpan` if you think its necessary.

frontend/src/app/components/ProfileStatistics/index.js
```diff
@@ -0,0 +1,79 @@
...
+              ))}
+            </tr>
+          ))}
+        </tbody>
+      </StatisticsTable>
+    </>
+  )
+}
+
+const createRowObject = (text, obj, sign, positive) => ({
```
littlewhywhat, line 61: My preference would be several separate simple components instead of one generic StatsRow with those three props creators... let's ask someone else as well. @novellizator @petrk-salsita what do you think?

frontend/src/app/components/ProfileStatistics/index.js
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <>
+      <TextSpan>Ranking: {ranking}</TextSpan>
+      <StatisticsTable>
+        <tbody>
+          {rows.map((row, index) => (
+            <tr key={index}>
+              {row.map((node, key) => (
+                <StatsRow key={[index, key].join('_')}
+                  text={node.text}
```
littlewhywhat, line 47: you can do also just <StatsRow {...node} /> to pass those, I think

frontend/src/app/components/ProfileStatistics/index.js
```diff
@@ -0,0 +1,79 @@
...
+      </StatisticsTable>
+    </>
+  )
+}
+
+const createRowObject = (text, obj, sign, positive) => ({
+  text,
+  player: obj.value,
+  score: `${obj.score}${sign}`,
+  positive: positive,
```
littlewhywhat, line 65: `let a; return { a: a }` could be `let a; return { a }`

frontend/src/app/components/ProfileStatistics/index.js
```diff
@@ -0,0 +1,79 @@
...
+  text,
+  altText: obj.value,
+  score: `${obj.score > 0 ? '+' : ''}${obj.score}`,
+  positive: obj.score > 0,
+})
+
+const createRankRowObject = (text, score, sign, positive) => ({
+  text,
+  score: `${sign}${score}`,
+  positive: positive,
```
littlewhywhat, line 78: same as above { positive }

frontend/src/app/modules/players/players-selectors.js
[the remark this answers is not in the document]
littlewhywhat: I think Dashboard and Leaderboard are using getTopPlayers, not getTopRatedPlayers, isn't it? that's why I asked because getTopRatedPlayers uses only ratings if I understand the code right. but even with criteria points - it could be that nextRank will show 0 if ratings are the same, right? that could be confusing...

frontend/src/app/modules/players/players-selectors.js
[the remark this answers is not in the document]
littlewhywhat: tricky one :) I realized that it happens because getRatingStatistics is undefined when you create a selector... try to define getRatingStatistics before creation of selector.

frontend/src/app/modules/players/players-selectors.js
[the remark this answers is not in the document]
littlewhywhat: Answered below if it was about undefined transform function in selector problem

frontend/src/app/modules/matches/matches-selectors.js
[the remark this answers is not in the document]
littlewhywhat: can't see what is changed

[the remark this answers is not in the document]
Fezzzi: Days,Teammates and Opponents are now precomputed and then extracted using `finxMax` and `findMin` functions.

frontend/src/app/modules/players/players-selectors.js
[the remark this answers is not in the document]
Fezzzi: Thanks, that did the thing.

frontend/src/app/modules/players/players-selectors.js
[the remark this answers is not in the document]
Fezzzi: Sorry, I meant conversation about [this](https://github.com/salsita/foosball-rating/pull/114/files/bf6e918f3cf0c320e1efcf81fab8d2d8d76e9025#diff-cd4220e6543e26c82876c33466c868e0) file but nevertheless, I read something about the problem and as I understand it there will be no such issue. There is only one instance of player statistics at the time and situation where multiple such would exist seems pretty exotic to me. Please correct me if I am wrong.

frontend/src/app/modules/matches/matches-selectors.js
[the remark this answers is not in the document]
littlewhywhat: yes but this selector will still have the problem as here https://github.com/salsita/foosball-rating/pull/114/files#r399260919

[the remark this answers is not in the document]
littlewhywhat: let's continue discussion in the other thread if needed - leaving that for reference

frontend/src/app/modules/players/players-selectors.js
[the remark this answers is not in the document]
Fezzzi: In case of default filters. there is no functional difference between `getTopPlayers` and `getTopRatedPlayers`. Yes, Leaderboard and Dashboard use the same component that uses `getTopPlayers`. The situation where someone will see value of `nextRank` or `prevRank` 0 will happen and is IMHO fine as it describes "The difference between your and the next palyer's score is...". It might be that the label's text does not say this correctly. I can either change label's text or make scores lower/higher by  to match actual label's text, what do you prefer?

frontend/src/app/components/ProfileStatistics/StatsRow.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+
+import { StatisticsValue, StyledLink, StatsSpan } from '../../../styles/blocks'
+import { createProfilePath } from '../../const/routes'
+
+export const StatsRow = ({ text, player, score, positive, altText = null }) => (
+  <td key="">{text}:
+    <StatisticsValue>
+      {player ? (
+        <StyledLink key={player?.id} to={createProfilePath(player?.id)}>
```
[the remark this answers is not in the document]
Fezzzi: Thanks good point.

frontend/src/app/components/ProfileStatistics/index.js
Fezzzi: I knew there was something like that but could get it to work, knowing only it was that easy... Thanks!

frontend/src/app/modules/players/players-selectors.js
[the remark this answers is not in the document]
littlewhywhat: agree it's exotic on this page, but if ProfileStatistics would be connected standalone component and could be used twice somewhere else then it would be the problem. For now it's ok then

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +77,83 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+
+export const computeOpponents = (playerId, playerMatches) =>
+  computePlayers(playerId, playerMatches, getOpponents) || {}
+
+const computePlayers = (playerId, playerMatches, playersProvider) => {
+  const playersMap = {}
+
+  for (const match of playerMatches) {
+    const players = playersProvider(playerId, match)
+    players.map(player =>
```
littlewhywhat, line 133: better to use forEach in this case than map as we are not mapping here anything...

frontend/src/app/components/ProfileStatistics/index.js
petrk-salsita: I agree that decomposing into reusable components would probably be cleaner 👍

frontend/src/app/modules/matches/matches-computations.js
petrk-salsita: Definitely a good point, in addition, `map` ideally shouldn't have side effects. `forEach` would be more suitable here

novellizator:
@Fezzzi would you like to rebase and finish this, or should we close it?

Fezzzi:
> @Fezzzi would you like to rebase and finish this, or should we close it?

I would still like to finish this as I think I have most of the remaining review comments done. I did not have time for that due to strict deadlines on the other project but I should have some time for that now or next week.

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Good idea, done [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-744525510377a0aca19d17cf4dc35116), [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-e1417bef467b606501c1db87ed6132c1) and below

frontend/src/app/components/ProfileStatistics/index.js
Fezzzi: Good idea, done [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-744525510377a0aca19d17cf4dc35116), [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-e1417bef467b606501c1db87ed6132c1) and below

Fezzzi:
Fixed links and the remaining CR points.

What is the reason behind these "new" snapshot tests @littlewhywhat @novellizator @petrk-salsita ? It seems that the only way to fix the one that fails is to rewrite it to match the new design that had to be denoted in order to fix the profile links. Wouldn't it be better to just remove this one particular test? It seems contraproductive to me as testing stuff that you either don't modify which implies the test passes or do modify, which implies you have to rewrite the tests to make it pass, does not yield much benefit.

However, my opinion might come from the lack of information.

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Done [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-304dca28d92d292252e86b7e83fc1569)

frontend/src/app/components/ProfileStatistics/index.js
Fezzzi: Changed the whole thing [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-744525510377a0aca19d17cf4dc35116)

frontend/src/app/components/ProfileStatistics/index.js
Fezzzi: Changed the whole thing [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-744525510377a0aca19d17cf4dc35116)

frontend/src/app/components/ProfileStatistics/index.js
Fezzzi: Changed the whole thing [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-744525510377a0aca19d17cf4dc35116)

frontend/src/app/components/ProfileStatistics/StatsRow.js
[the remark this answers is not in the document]
Fezzzi: I don't have to check for `player` existence at all after breaking into separate "specialized" components. [here](https://github.com/salsita/foosball-rating/pull/114/commits/48d97a59bd3178627233fc0798f8d61a47a09c3b#diff-5901ef1a93d3481680fe5cf020ca8ea7)

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +79,71 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+export const findMax = (arr, getField = node => node.score) =>
+  findWithComparator((e1, e2) => e1 >= e2, arr, getField, -Infinity)
+
+export const computeDays = matchChanges => {
+  const daysMap = new Map()
+
+  for (const match of matchChanges) {
+    const day = match.date.toLocaleDateString()
+    const score = Number(match.ratingChangeString)
+    daysMap.set(day, !daysMap.has(day)
```
littlewhywhat, line 105: it's common practice to minimize use of the negations as it's easier to read... if you just turn this condition as `daysMap.has(day)? A : B` it could be easier to read than `!daysMap.has(day)? B : A`

littlewhywhat: otherwise I like the use of Map here :+1:

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +79,71 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+export const computeOpponentsStats = (playerId, playerMatches) =>
+  computePlayerStats(playerId, playerMatches, getOpponents) || {}
+
+const computePlayerStats = (playerId, playerMatches, playersProvider) => {
+  const playersMap = new Map()
+
+  for (const match of playerMatches) {
+    const players = playersProvider(playerId, match)
+    players.forEach(player =>
+      playersMap.set(player.id, !playersMap.has(player.id)
```
littlewhywhat, line 132: i see that you are fond of negating first :D well I guess there are two kinds of people in the world :D

frontend/src/app/modules/players/players-selectors.js
```diff
@@ -62,3 +62,22 @@ const getPlayerCriteriaPoints = (player, matchesLast, criteria) => {
...
 }
+
+const getRatingStatisticsForPlayer = (players, playerId) => {
+  const index = players.findIndex(player => player.id === playerId)
+  return {
+    ranking: index + 1,
+    scoreToNextRank: index > 0
+      ? 1 + (players[index - 1].rating - players[index].rating)
+      : 0,
+    scoreToPevRank: index < players.length - 1
```
littlewhywhat, line 73: `scoreToPrevRank`?

littlewhywhat:
> It seems that the only way to fix the one that fails is to rewrite it to match the new design that had to be denoted in order to fix the profile links. Wouldn't it be better to just remove this one particular test? It seems contraproductive to me as testing stuff that you either don't modify which implies the test passes or do modify, which implies you have to rewrite the tests to make it pass, does not yield much benefit.

Well that sounds for me as a generic description for any unit test :D... However the benefit is to "feel the change"... not sure which snapshot you mean in particular though...
 - usually we test branching - different renders on the different conditions... therefore if you change something in one branch, you know that other branches are still correct. I think benefit here is clear.
 - if there is no branching it's still can benefit because you can change something in the component that shouldn't affect how it renders in your opinion but it does somehow and snapshot will show you.

Overhead is not so big - eventually just hit update and then smth weird could be seen on the review. Of course, there are trade offs and sometimes snapshots could be overkill I guess too.

In summary, unit tests tend to fail and that's kind of their purpose in my opinion :man_shrugging:

frontend/src/app/Game.js
```diff
@@ -49,7 +49,9 @@ export class GameComponent extends Component {
               <Route exact path={constructUrl(ROUTES.MATCH_LIST)}>
                 <MatchListPage constructUrl={constructUrl} />
               </Route>
-              <Route exact path={constructUrl(ROUTES.PROFILE)} component={Profile} />
+              <Route exact path={constructUrl(ROUTES.PROFILE)} render={matchProps => (
```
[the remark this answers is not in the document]
Fezzzi: We need to pass those `matchProps` to Profile component and I don't know of other way to pass url matched by the `Route` to the component aside from this.

frontend/src/app/modules/players/players-selectors.js
Fezzzi: Thanks, [fixed](https://github.com/salsita/foosball-rating/pull/114/commits/54023373857edbdd2d03d427c0b799b98214f08c#diff-744525510377a0aca19d17cf4dc35116)

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: De-negated [here](https://github.com/salsita/foosball-rating/pull/114/commits/54023373857edbdd2d03d427c0b799b98214f08c#diff-304dca28d92d292252e86b7e83fc1569)

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: De-negated [here](https://github.com/salsita/foosball-rating/pull/114/commits/54023373857edbdd2d03d427c0b799b98214f08c#diff-304dca28d92d292252e86b7e83fc1569)

Fezzzi:
Yeah, I see there are some benefits of this. I fixed the tests with edditing [this](https://github.com/salsita/foosball-rating/pull/114/commits/10d6668f1e7d8926d953be72210cb677cd524cd3#diff-65dafe036251aced9cebd33024b472b8) change, snapshot.

> In summary, unit tests tend to fail and that's kind of their purpose in my opinion 🤷‍♂️

I still cannot agree with this though. Tests should IMO point out errors and unintentional behaviour changes but there shouldn't be really need to modify them. It might be only because I have never worked with this kind of tests before, but generally, you unit-test some unit of fuctionality regardless of the actual code bellow so changing the way it is implemented does not break the tests as long as it behaves the same as the code before in all the required test cases + you define new tests to unsure preserving of additional behaviour your new implementation yields. All in all, I would ban using snapshot testing 😀

frontend/src/app/Game.js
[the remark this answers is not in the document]
littlewhywhat: so for the record the difference from other cases here is that we need to pass those Router props and constructUrl together... having it as a child we won't get the first and putting it as a component prop we won't get the second... Thus the `render` is used :)

littlewhywhat:
> you unit-test some unit of fuctionality regardless of the actual code bellow so changing the way it is implemented does not break the tests as long as it behaves the same as the code before in all the required test cases + you define new tests to unsure preserving of additional behaviour your new implementation yields. All in all, I would ban using snapshot testing

what you describe sounds more like black box testing while unit tests (even though sometimes compromised) are actually white box tests... hence code coverage makes sense. Implementing "all required test cases" is covering all branches - in most cases change will require unit tests changes as well in this case.

Unit tests bring with themselves, of course, a burden of maintenance - that's why many developers are often reluctant to use them :) But they also bring confidence that here and there code works as it did before when you try to localize the bug. Also there are often problems with auto-merges.

Say Game component failed to render but test with snapshot passes... that gives me a clue that the problem is somewhere else ;)

anyway get approval from @novellizator @petrk-salsita and let's merge this PR! :D

PR #115, opened 2020-03-13
Issue 11 top rating user title

### LINKED ISSUE #11: Improvement: Top Ratings
For King of the Hill user (Profile who have biggest ELO rating) ad countdown how long he have biggest ELO rating.  (  Jhon 2840 (KotH last 180 day's)

Fezzzi:
Closes #11

### DIFF
--- a/frontend/src/app/modules/matches/matches-computations.js
+++ b/frontend/src/app/modules/matches/matches-computations.js
@@ -80,27 +80,27 @@ export const computeKingStreakDuration = (matchesLast, usersLast) => {
   }, new Map())
 
   const kingId = usersLast[0].id
-  let matchFound = false;
-  for (let match of matchesLast) {
+  let matchFound = false
+  for (const match of matchesLast) {
     const players = [...match.team1, ...match.team2]
-    
+
     // #1 recomptute king's rating before in case he played the match
     const kingPlayer = players.find(player => player.id === kingId)
-    const kingWon = false;
+    let kingWon = false
     if (kingPlayer) {
       kingWon = usersMap[kingId] > kingPlayer.matchRating
       usersMap[kingId] = kingPlayer.matchRating
     }
 
     // #2 check whether none of the other players beat the king
-    for (let player of players) {
+    for (const player of players) {
       usersMap[player.id] = player.matchRating
       matchFound |= (player.id !== kingId && player.matchRating > usersMap[kingId])
     }
 
     // #3 check if king was first before winning
     if (kingPlayer && kingWon && !matchFound) {
-      for (let key in usersMap) {
+      for (const key in usersMap) {
         matchFound |= (usersMap[key] > usersMap[kingId])
       }
     }
--- a/frontend/src/app/modules/matches/matches-computations.js
+++ b/frontend/src/app/modules/matches/matches-computations.js
@@ -68,3 +68,49 @@ export const computeWinRatio = (playerId, playerMatches) => {
   const wonMatchesCount = playerMatches.filter(match => didPlayerWin(playerId, match)).length
   return (playerMatches.length > 0) ? (wonMatchesCount / playerMatches.length) : 0
 }
+
+export const computeKingStreakDuration = (userId, users, matches) => {
+  // todo: remove usersLast and matchesLast when it is not needed after merging leaderboards branch
+  const usersLast = users.sort((user1, user2) => user2.rating - user1.rating)
+  const matchesLast = matches.sort((match1, match2) => match2.date - match1.date)
+  if (!usersLast.length || !matchesLast.length || usersLast[0].id !== userId) {
+    return false
+  }
+
+  const usersMap = usersLast.reduce((map, user) => {
+    map[user.id] = user.rating
+    return map
+  })
+
+  let kingRating = usersLast[0].rating
+  for (let i = 0; i < matchesLast.length; ++i) {
+    const players = [...matchesLast[i].team1, ...matchesLast[i].team2]
+    
+    // #1 recomptute king's rating before in case he played the match
+    const kingPlayer = players.find(player => player.id === userId)
+    const kingWon = false;
+    if (kingPlayer) {
+      kingWon = kingRating > kingPlayer.matchRating
+      kingRating = kingPlayer.matchRating
+    }
+
+    // #2 check whether none of the other players beat the king
+    for (let player of players) {
+      usersMap[player.id] = player.matchRating
+      if (player.id !== userId && player.matchRating > kingRating) {
+        return matchesLast[i]
+      }
+    }
+
+    // todo: remove `key < 180` when it is not needed after merging leaderboards branch
+    // #3 check if king was first before winning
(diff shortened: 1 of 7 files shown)


Fezzzi:
There are some issues with loading data, I will fix that with rebase after merging #27 so lets focus on that for now and ignore this one...

frontend/src/app/components/BattleHistory/BattleHistoryRow.js
```diff
@@ -6,7 +6,7 @@ import {
 import { Team } from './Team'
 
 export const BattleHistoryRow = ({ match }) =>
-  <ListItem Display="grid" Column="2fr 1fr 2fr">
+  <ListItem display="grid" column="2fr 1fr 2fr">
```
Fezzzi, line 9: Just for the sake of consistency

frontend/src/app/modules/matches/matches-selectors.js
```diff
@@ -52,6 +51,6 @@ export const getStatisticsForPlayer = createSelector(
 
 export const getRatingHistoryGraphForPlayer = createSelector(
   getLastMatchesForPlayer,
-  getPlayer,
+  (state, playerId) => state.players.find(player => player.id === playerId),
```
Fezzzi, line 54: I had to change this due to (probably) circular dependency in selectors between `matches-selectors` and `players-selectors`:
```
Error: Selector creators expect all input-selectors to be functions, instead received the following types: [function, undefined]
```
An ideas about alternative solution are welcome!

frontend/src/styles/blocks/list.js
```diff
@@ -48,12 +48,13 @@ const ListCon = styled.ul`
 
 const ListItem = styled.li`
   padding: ${variables.baseSpacing};
-  display: ${props => props.Display};
-  grid-template-columns: ${props => props.Column};
+  display: ${props => props.display};
+  grid-template-columns: ${props => props.column};
```
Fezzzi, line 52: Just for the sake of consistency

frontend/src/app/modules/matches/matches-selectors.js
littlewhywhat: usually the solution for A <-> B is to add C :) so A <- C, B <- C :) so I guess some selectors could be put outside of matches/players ones

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +68,49 @@ export const computeWinRatio = (playerId, playerMatches) => {
   const wonMatchesCount = playerMatches.filter(match => didPlayerWin(playerId, match)).length
   return (playerMatches.length > 0) ? (wonMatchesCount / playerMatches.length) : 0
 }
+
+export const computeKingStreakDuration = (matchesLast, playersLast) => {
```
littlewhywhat, line 72: matchesLast -> lastMatches, playersLast -> topPlayers?

frontend/src/app/modules/matches/matches-computations.js
```diff
@@ -68,3 +68,49 @@ export const computeWinRatio = (playerId, playerMatches) => {
...
+  }
+
+  const playersMap = playersLast.reduce((map, player) => {
+    map[player.id] = player.rating
+    return map
+  }, new Map())
+
+  const kingId = playersLast[0].id
+  let matchFound = false
+  for (const match of matchesLast) {
```
littlewhywhat, line 84: looks a bit like black magic to me :D i guess idea is to go backwards until we find some match (moment in time) where the king became a king, right?

littlewhywhat: maybe if you could a bit elaborate on those 3 steps... The way i see how it can be computed is smth like that:
1. King may become a such by playing a match or by somebody else playing a match (and losing rating)
2. I think we have a snapshot of players ratings in their matches
3. If King became a such by other losing some rating then it seems easy - just find a match where someone has a bigger rating then King's. I guess it's right? :man_shrugging: 
4. If King became a such by gaining some rating then we need somehow to realize that the previous King has lower rating... Previous King can be or not in the match - so I am afraid we need to compare somehow to all other players here (or at least the second biggest at that moment - has to be computed somehow) and understand if this won match is this deal-breaker match...

Probably first 3 I somehow see in the code below but definitely missing 4th point... maybe I am missing something in my conclusions?...

littlewhywhat: in any case not a trivial algorithm definitely needs more docs/comments or maybe we should just solve it by adding something simple in the BE instead

littlewhywhat: @petrk-salsita @novellizator any thoughts? :)

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: Renamed [here](https://github.com/salsita/foosball-rating/pull/115/commits/386ab96fc278ec02496779a90c5d94567534cb46#diff-e6a7cdc081548e43f8be9112e29eb70d)

frontend/src/app/modules/matches/matches-selectors.js
Fezzzi: I thought about it for a bit and decided that conceptually, there is no need for third group of selectors as they are all still selectors of players and matches. I tried to reorganize and refactor selectors and computations but it is hard to decide when a thing working with both players and matches should be labeled as `players-selector` and when as `matches-selector`. I used as a rule of a thumb some similaties (such as all `getXXXForPlayer` selectors are together) and general functionalities of things (like `computeKingStreakDuration` seems more of players-computation rather then matches one).

What do you think about the current design and what would you improve/change @littlewhywhat ?

littlewhywhat: if it works and there is no unnecessary code duplication (it can be accepted sometimes though), I am fine with it :) I was mainly answering your question) I don't mind that much about organizing things on a such low level... I think in this case, one (or at least me) would use IDE tools to navigate through those function anyway. And there are probably hundreds of way to organize these functions and none of them is probably perfect :man_shrugging: so if it makes sense to you it's good :+1:

frontend/src/app/modules/matches/matches-computations.js
Fezzzi: I added some brief explanation of the algorithm and it here as well:
```
  First, we create a hash table of players and their current scores
  Then we iterate through matches from the latest, recompute scores of affected players and compare
  More preciselly:
    1. Recompute king's score in case he played the match, note whether he won the match
    2. Recompute scores of all (other) players in the match and test if any of them tops king
    3. Otherwise, compare his new score with scores of all other players,
    in case the king played the match and won (thus had smaller score before).
```
How you described it is basically the way it is implemented, looping over players affected by the match (your 3.) is 2. in the docs (lines 84-87) and looping over all players in case King won the match (your 4.) is 3. in the docs (lines 89-93)

littlewhywhat (approved):
lgtm

PR #117, opened 2020-03-13
gh-116: Let UI use games api

### LINKED ISSUE #116: Start using Games API from UI
make the UI to use new API - and hardcode for now that the first game type is used (which is foosball)

- [x] fetch all game types from BE and select first one as current
- [x] add new match with the current game type id
- OPTIONAL: (could be done later) fetch matches only based on current game type id - requires another api point in BE - will be done in #118

littlewhywhat:
After #112 
Closes #116 

Add /games/:name/matches endpoint and use it to store the match related to a particular game. Add sagas to fetch games and select a game. Make fetching of matches wait until the game is selected.

### DIFF
--- a/backend/repositories/MatchRepository.ts
+++ b/backend/repositories/MatchRepository.ts
@@ -15,7 +15,7 @@ const updateRatingForTeam = async (team, difference, storageContext): Promise<vo
   }))
 }
 
-const storeMatch = async (match): Promise<MatchWithId> => {
+const storeMatch = async (match: Match): Promise<MatchWithId> => {
   const winningTeam = match.team1Won ? match.team1 : match.team2
   const losingTeam = match.team1Won ? match.team2 : match.team1
 
@@ -56,7 +56,8 @@ const getRatingChanges = ({ team1, team2 }, team1Won): RatingChanges => {
   return ratingCalculator.computeRatingChanges(winningTeam, losingTeam)
 }
 
-const constructMatch = async (matchDescription): Promise<Match> => {
+const constructMatch = async (gameName: string, matchDescription): Promise<Match> => {
+  const game = await storage.getGameByName(gameName)
   const teams = {
     team1: await getFilledTeam(matchDescription.team1),
     team2: await getFilledTeam(matchDescription.team2),
@@ -71,7 +72,8 @@ const constructMatch = async (matchDescription): Promise<Match> => {
     matchDescription.team1Won,
     date,
     winningTeamRatingChange,
-    losingTeamRatingChange
+    losingTeamRatingChange,
+    game.id
   )
 }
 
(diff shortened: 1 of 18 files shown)


backend/repositories/MatchRepository.ts
```diff
@@ -92,13 +94,13 @@ const getElapsedSecondsSinceLatestMatch = async (): Promise<number> => {
   * @param {Array<number>} matchDescription.team2 Array of 1 or 2 elements containing IDs of players from team2.
   * @param {boolean} matchDescription.team1Won True if team1 won, false if team2 won.
   */
-export const recordMatch = async (matchDescription): Promise<Match> => {
+export const recordMatch = async (gameName: string, matchDescription): Promise<Match> => {
```
novellizator: i'd add an interface/type for matchDescription, and also add check isMatchDescription, like did in the friday's PR

frontend/src/app/modules/games/games-actions.js
```diff
@@ -0,0 +1,8 @@
+import { createActions } from 'reduxsauce'
+
+export const GamesActions = createActions({
+  gamesLoaded: ['games'],
+  selectGame: ['selectedGame'],
+}, {
+  prefix: 'foosball-rating/games/',
```
novellizator: small issue - I might consider removing the "foosball-rating" prefix, as we're considering to rename the application and you don't want to leave traces like that ;)

littlewhywhat: wouldn't it be hard to find where you need to rename/add prefix unless you have each place with an old name?

frontend/src/app/modules/games/games-saga.test.js
```diff
@@ -0,0 +1,29 @@
+import { getGames } from './games-effects'
+import { getGamesSaga } from './games-saga'
+import { call, put } from 'redux-saga/effects'
+import { GamesActions } from './games-actions'
+
+const RESPONSE = {
+  data: {},
+}
+
+describe('gameSaga', () => {
```
novellizator: fyi: I've seen many projects NOT test sagas, as they didn't see the clear benefit to it. hopefully, after your tests they will shine through

littlewhywhat: I think it's a common misconception that tests are needed for something complex only. The main benefit that I experienced is that once you do some refactoring or renaming or automerge some nasty things can happen even in a simple code and unit tests can guard against it as they will feel it :)

backend/repositories/MatchRepository.ts
littlewhywhat: it is already added here https://github.com/salsita/foosball-rating/pull/129/commits/69717c292c04a2fac323ec8cb31289dfdbdbf8b8 and I will adjust it as in the friday's PR. Is it ok with you?

novellizator: fine with me!

Issue #118, opened 2020-03-13
Make UI fetch matches by game type id


littlewhywhat:
- [x] extend BE api for games to return related matches
- [x] adjust sagas

PR #125, opened 2020-03-17
gh-118: Use api games/:name/matches from UI

littlewhywhat:
Closes gh-118
after #116

### DIFF
--- a/frontend/src/app/modules/games/games-saga.js
+++ b/frontend/src/app/modules/games/games-saga.js
@@ -18,6 +18,7 @@ export function* selectGameSaga(game) {
 }
 
 export function* selectGameOnGamesLoadedSaga(action) {
+  // select foosball as first game as it's the only game right now
   yield call(selectGameSaga, action.games[0])
 }
 
--- a/frontend/src/app/modules/games/games-saga.js
+++ b/frontend/src/app/modules/games/games-saga.js
@@ -1,6 +1,8 @@
-import { put, fork, call, takeEvery } from 'redux-saga/effects'
+import { put, fork, call, takeEvery, select } from 'redux-saga/effects'
 import { GamesActions } from './games-actions'
 import { getGames } from './games-effects'
+import { MatchesActions } from '../matches/matches-actions'
+import { getSelectedGame } from './games-selectors'
 
 export function* getGamesSaga() {
   try {
@@ -11,11 +13,21 @@ export function* getGamesSaga() {
   }
 }
 
-export function* selectGameSaga(action) {
-  yield put(GamesActions.Creators.selectGame(action.games[0]))
+export function* selectGameSaga(game) {
+  yield put(GamesActions.Creators.selectGame(game))
+}
+
+export function* selectGameOnGamesLoadedSaga(action) {
+  yield call(selectGameSaga, action.games[0])
+}
+
+export function* refreshSelectionSaga() {
+  const selectedGame = yield select(getSelectedGame)
+  yield call(selectGameSaga, selectedGame)
 }
 
 export function* gamesSaga() {
   yield fork(getGamesSaga)
-  yield takeEvery(GamesActions.Types.GAMES_LOADED, selectGameSaga)
+  yield takeEvery(GamesActions.Types.GAMES_LOADED, selectGameOnGamesLoadedSaga)
+  yield takeEvery(MatchesActions.Types.MATCH_ADDED, refreshSelectionSaga)
 }
--- a/frontend/src/app/modules/matches/matches-saga.test.js
+++ b/frontend/src/app/modules/matches/matches-saga.test.js
@@ -15,11 +15,11 @@ const ERROR = {
 describe('matchesSaga', () => {
   describe('addMatchSaga', () => {
     let gen
-    const expectNext = (suppliedValue, generated) => {
-      expect(gen.next(suppliedValue)).toStrictEqual(generated)
+    const expectNext = (suppliedValue, expected) => {
+      expect(gen.next(suppliedValue)).toStrictEqual(expected)
     }
-    const expectThrow = (suppliedValue, generated) => {
-      expect(gen.throw(suppliedValue)).toStrictEqual(generated)
+    const expectThrow = (suppliedValue, expected) => {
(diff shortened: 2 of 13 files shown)


frontend/src/app/modules/games/games-saga.js
```diff
@@ -11,11 +13,21 @@ export function* getGamesSaga() {
...
+
+export function* refreshSelectionSaga() {
+  const selectedGame = yield select(getSelectedGame)
+  yield call(selectGameSaga, selectedGame)
 }
 
 export function* gamesSaga() {
   yield fork(getGamesSaga)
-  yield takeEvery(GamesActions.Types.GAMES_LOADED, selectGameSaga)
+  yield takeEvery(GamesActions.Types.GAMES_LOADED, selectGameOnGamesLoadedSaga)
```
novellizator: this is a matter of taste, feel free to keep it like it is. I'd probably consider different names to these sagas, namely onGamesLoadedSaga and onMatchAddedSaga. But this is fine as well.

frontend/src/app/modules/games/games-saga.js
```diff
@@ -11,11 +13,21 @@ export function* getGamesSaga() {
...
 }
 
-export function* selectGameSaga(action) {
-  yield put(GamesActions.Creators.selectGame(action.games[0]))
+export function* selectGameSaga(game) {
+  yield put(GamesActions.Creators.selectGame(game))
+}
+
+export function* selectGameOnGamesLoadedSaga(action) {
+  yield call(selectGameSaga, action.games[0])
```
novellizator: what is that games[0] referring to? 🤔

frontend/src/app/modules/games/games-saga.js
littlewhywhat: can't we put another saga on the same Action?

frontend/src/app/modules/games/games-saga.js
littlewhywhat: to foosball :) as right now we "can't" have more games, I just take the first one of one existing.

littlewhywhat: it's temporary as long as we support only one game on UI

frontend/src/app/modules/matches/matches-saga.test.js
```diff
@@ -0,0 +1,55 @@
...
+const MATCH = {}
+const FOOSBALL = { name: 'foosball' }
+const ERROR = {
+  message: 'ERROR',
+}
+
+describe('matchesSaga', () => {
+  describe('addMatchSaga', () => {
+    let gen
+    const expectNextToBe = (value, done = false) => {
```
novellizator: nice little helper ;) As a idea: since the unit tests here serve strictly as a regression. I'd consider snapshot test this, to save the hassle. However, this is fine as well!

frontend/src/app/modules/matches/matches-saga.test.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+    beforeEach(() => {
+      gen = addMatchSaga({
+        match: MATCH,
+      })
+    })
+    describe('when addMatch is successful', () => {
+      it('performs correctly the sequence', () => {
+        expectNextToBe(select(getSelectedGame))
+        expect(gen.next(FOOSBALL).value)
```
novellizator: You could theoretically use expectNextToBe(expectedResult, suppliedValue=undefined) here.
where expectedResult could be an object having optional `value` and `done` properties which would be check in case of existence

frontend/src/app/modules/games/games-saga.js
novellizator: Theoretically we can; practically I found it can get tricky to take the same action from multiple places as there might be concurrency issues and also both sagas could to some overlapping operation (both are a result on bad architecture of course). However, like it said, this is fine as well.

frontend/src/app/modules/games/games-saga.js
novellizator: if it's temporary I'd either make a comment here, or rename the saga to something like selectFoosball - so the name says what it does

frontend/src/app/modules/games/games-saga.js
littlewhywhat: ok, let's leave as it is but I see your point

frontend/src/app/modules/matches/matches-saga.test.js
littlewhywhat: agreed to leave as it is

frontend/src/app/modules/games/games-saga.js
littlewhywhat: Fixed in https://github.com/salsita/foosball-rating/pull/125/commits/87707178781b189c0157f2314e3bb531c773c7c0

frontend/src/app/modules/matches/matches-saga.test.js
littlewhywhat: Fixed in https://github.com/salsita/foosball-rating/pull/125/commits/87707178781b189c0157f2314e3bb531c773c7c0

littlewhywhat: I merged yours and mine suggestions :)

frontend/src/app/modules/matches/matches-saga.test.js
```diff
@@ -15,13 +15,28 @@ const ERROR = {
 describe('matchesSaga', () => {
   describe('addMatchSaga', () => {
     let gen
-    const expectNextToBe = (value, done = false) => {
-      const result = gen.next()
-      expect(result.value).toStrictEqual(value)
-      expect(result.done).toBe(done)
+    const expectNext = (suppliedValue, generated) => {
+      expect(gen.next(suppliedValue)).toStrictEqual(generated)
```
novellizator: i'd call this "expected" not generated

littlewhywhat: agree, thanks, fixed in https://github.com/salsita/foosball-rating/pull/125/commits/4a803ce91ad0348134a945f4c581076d76e339aa

PR #129, opened 2020-03-19
gh-126: Separate ratings to Players table

### LINKED ISSUE #126: Add Players table
add Player table with Initial, Current ratings, GameId, UserId - use it to store ratings but preserve API as it is currently

littlewhywhat:
after #118 
Closes #126


backend/app.ts
```diff
@@ -43,7 +44,7 @@ const processError = (response, error): void => {
 }
 
 app.get('/users', (req, res) => {
```
novellizator: maybe rename the endpoint to /players? 🤔

littlewhywhat: I will handle renaming in future PR - I left it like this for now to avoid changes in FE.

backend/db_migrations/1584979214020_add-no-duplicate-players-in-matches.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.addConstraint('Matches', 'NO_DUPLICATE_PLAYERS_IN_TEAM_1', {
+    check: '"Team1Player1Id" != "Team1Player2Id"',
+  })
+  pgm.addConstraint('Matches', 'NO_DUPLICATE_PLAYERS_IN_TEAM_2', {
+    check: '"Team2Player1Id" != "Team2Player2Id"',
+  })
+  pgm.addConstraint('Matches', 'NO_DUPLICATE_PLAYERS_IN_MATCHES', {
+    check: '"Team1Player1Id" != "Team2Player1Id" AND ' +
```
novellizator: i'd use backticks ` , so you don't need to concatenate the strings

littlewhywhat: I used this way for easier formatting and at the same time not creating unnecessary whitespaces in the constraint definition. with backsticks and if I keep those on separate lines there will spaces from the beginning of the line if I understand it correctly

backend/db_migrations/1584981130161_add-players-have-same-game-id-as-match.sql
```diff
@@ -0,0 +1,23 @@
+CREATE FUNCTION MATCHES_AFTER_INSERT_TRIGGER_CREATE()
+RETURNS trigger
+LANGUAGE PLPGSQL AS
+$$BEGIN
+  IF EXISTS (
+    SELECT * FROM "Matches" m, "Players" p
```
novellizator: can't you use inner join between matches and players here? 🤔

littlewhywhat: it is an inner join :relaxed: https://www.postgresql.org/docs/8.3/tutorial-join.html

backend/db_migrations/1584981130161_add-players-have-same-game-id-as-match.sql
```diff
@@ -0,0 +1,23 @@
...
+    AND p."GameId" <> NEW."GameId"
+  )
+  THEN
+    RAISE EXCEPTION 'A player in a match cant play a different game than others';
+  END IF;
+  RETURN NEW;
+END;$$;
+
+CREATE TRIGGER MATCHES_AFTER_INSERT_TRIGGER AFTER INSERT ON "Matches"
+FOR EACH ROW EXECUTE PROCEDURE MATCHES_AFTER_INSERT_TRIGGER_CREATE();
```
novellizator: do we need to run on on all rows, or only on the new added one?

littlewhywhat:
fixed team ids as integers from UI in https://github.com/salsita/foosball-rating/pull/129/commits/be48ff0b2519e426d4045eac1386c3f61034db4a

backend/db_migrations/1584981130161_add-players-have-same-game-id-as-match.sql
littlewhywhat: it's 'FOR EACH' added row - https://www.postgresql.org/docs/9.1/sql-createtrigger.html

backend/db_migrations/1584979214020_add-no-duplicate-players-in-matches.ts
novellizator: IMO, you don't have to worry about that - if you add a backslash at the end of the line, it doesn't add a new line. Theoretically you can use tagged templates (https://muffinresearch.co.uk/removing-leading-whitespace-in-es6-template-strings/) to remove the leading whitespaces.

I'm not going to impose the backtick style, you can keep this, but I personally strongly prefer readability over a few added whitespaces (which might be trimmed somewhere down the road)

backend/db_migrations/1584981130161_add-players-have-same-game-id-as-match.sql
novellizator: yes, I know. I'm suggesting to use the dedicated syntax for it, so the intent is even more clear

littlewhywhat:
add verbose errors for invalid match data in https://github.com/salsita/foosball-rating/pull/129/commits/b10c1b0c2866ff8f24eee3f1285732d38f38aa5b

backend/db_migrations/1584981130161_add-players-have-same-game-id-as-match.sql
littlewhywhat: no join anymore :) https://github.com/salsita/foosball-rating/pull/129/commits/2d493ce517f94386c34244ee51f814d41ee6896b

backend/db_migrations/1584979214020_add-no-duplicate-players-in-matches.ts
littlewhywhat: cool, didn't know about that! fixed with common-tags here https://github.com/salsita/foosball-rating/pull/129/commits/1f5c04659da17d5c5fa7e47053799e439a44ca71

littlewhywhat:
User creation was broken - fixed here https://github.com/salsita/foosball-rating/pull/129/commits/70c6601923d749f8af9cd03e9b955d6521a1deff

PR #141, opened 2020-03-26
gh-132: add Game switch

### LINKED ISSUE #132: Use router to switch between games
- [x] Make nested routes with game fixed to 'foosball'
- [x] Implement a game switch via URL

littlewhywhat:
after #139 
closes #132 
after #143


littlewhywhat:
- [ ] currently selection happens on each AppComponent update - probably if gameName matches current selection then we should not reselect it from AppComponent

littlewhywhat:
Put unit tests in a separate commit https://github.com/salsita/foosball-rating/pull/141/commits/7915cb441ece84208a8ba850de3da4cb981761c3 - probably can simplify review

littlewhywhat:
for some reason there is no data in deployment - trying to see how it will be without test data extension

frontend/src/Root.js
```diff
@@ -16,16 +19,19 @@ const store = createStore(rootReducer,
...
   <Provider store={store}>
     <Theme>
-      <Router>
+      <Router history={history}>
         <Switch>
-          <Route path='/foosball' component={App}/>
-          <Route>
-            <Redirect to='/foosball'/>
-          </Route>
+          <Route path='/:gameName' component={App}/>
```
novellizator, line 33: i'd consider renaming App so that the're no confusion between Root and App.
Say `<SpecificGameType>, <GameType>, <Game>` ?

frontend/src/app/App.js
```diff
@@ -12,40 +12,52 @@ import { CreateMatchPage } from './pages/CreateMatch'
...
         </Container>
         <Footer />
       </>
     )
   }
 }
+
+const mapStateToProps = state => ({
+  isGameSelected: isGameSelected(state),
+  gameNotFound: state.gameNotFound,
```
novellizator, line 60: 2 questions: 
1. isn't gameNotFound complementary to isGameSelected? Or can you select games that are not found?
2. usually we don't sue negative in variable names. it's easier to understamd variable isGameFound that gameNotFound

frontend/src/app/App.js
```diff
@@ -12,40 +12,52 @@ import { CreateMatchPage } from './pages/CreateMatch'
...
+import { isGameSelected } from './modules/games/games-selectors'
+import { connect } from 'react-redux'
 
-export class App extends Component {
+export class AppComponent extends Component {
   render() {
-    const { match: { url } } = this.props
+    const { match: { url, params: { gameName } }, isGameSelected, gameNotFound } = this.props
     const createMatch = () => {
       this.props.history.push(`${url}${ROUTES.CREATE_MATCH}`)
```
novellizator, line 23: i know it's not directly related to this PR, but this is kind of an antipattern - because you never know if there's exactly one backslash in url, or routes.creatematch. usually a url is constructed by joining two paths by some third party library that takes care of this, something like url-join etc.
Not a big issue though

frontend/src/app/App.js
littlewhywhat: 1. the game may not be selected and may be still found though... gameNotFound as true means that the selection failed...

littlewhywhat: 2. because of 1 it would be more confusing because isGameFound true doesn't necessary mean that it actually is found but could be that it's was not yet selected at all... probably the name is wrong at all

frontend/src/Root.js
littlewhywhat: Let's call it Game :)

frontend/src/app/App.js
littlewhywhat: fixed with enum here https://github.com/salsita/foosball-rating/pull/141/commits/8c9de1db3591e4a5cb8bbdf72d80fe0052794850

frontend/src/Root.js
littlewhywhat: fixed in https://github.com/salsita/foosball-rating/pull/141/commits/d00b12891dc1735349c2f42b03bbfc42f1f68aa4

frontend/src/app/App.js
littlewhywhat: Add url-join in https://github.com/salsita/foosball-rating/pull/141/commits/65b89d580928f1960523b96e1fa2c2e72be643b5

littlewhywhat: this actually breaks the addMatch redirect completely because of the bug that is solved here https://github.com/salsita/foosball-rating/pull/150

frontend/src/app/App.js
littlewhywhat: reverted and fixed with https://github.com/salsita/foosball-rating/pull/141/commits/04c18104046efea46d420642ac764d846b2c69a9

Issue #153, opened 2020-04-06
Add Github action for automation


littlewhywhat:
Create a basic test with cypress and add it as Github action

littlewhywhat:
closed here https://github.com/salsita/foosball-rating/pull/154

PR #154, opened 2020-04-06
gh-153: First simple e2e test

littlewhywhat:


package.json
```diff
@@ -5,8 +5,10 @@
   "main": "index.js",
   "scripts": {
     "start": "cd backend && npm run create-schema && cd .. && ./start.sh",
-    "postinstall": "cd backend && npm install && cd ../frontend && npm install && npm run build && cd ..",
-    "lint": "eslint --max-warnings 0 './backend/**/*.ts' './frontend/src/**/*.js' --ignore-pattern '/backend/node_modules/*' --ignore-pattern '/frontend/node_modules/*'"
+    "lint": "eslint --max-warnings 0 './backend/**/*.ts' './frontend/src/**/*.js' --ignore-pattern '/backend/node_modules/*' --ignore-pattern '/frontend/node_modules/*'",
```
novellizator, line 8: can't you simply ignore the node_modules altogether without listing the out in both directories?

package.json
```diff
@@ -5,8 +5,10 @@
...
+    "start:e2e": "concurrently \"cd backend && npm start\" \"wait-on http://localhost:3000/games && cd frontend && npm run start:e2e\"",
```
novellizator, line 9: do we need to specifically wait on "/games" endpoint, or is "/" enough?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  e2e:
+    runs-on: ubuntu-latest
+    # cypress requires xvfb
+    container: thompsnm/nodejs-chrome-xvfb
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_PASSWORD: XXXX
```
novellizator, line 18: i might make the user and db explicit (POSTGRES_USER, POSTGRES_DB)

.github/workflows/build.yml
```diff
@@ -17,10 +17,12 @@ jobs:
     - name: FE unit tests
       run: |
         cd frontend
+        npm install
```
novellizator, line 20: you might want to take a look at this https://docs.npmjs.com/cli/ci.html

package.json
littlewhywhat: true, thank you, fixed here https://github.com/salsita/foosball-rating/pull/154/commits/6de4130a5334e3f7cc77b41a9f0f53a38de58ed2

package.json
littlewhywhat: wait-on waits for 2xx on GET and for "/" we give 404 only right now.

.github/workflows/build.yml
littlewhywhat: thanks, fixed here https://github.com/salsita/foosball-rating/pull/154/commits/cd851298d2d5058b4f9b94c5e69b4e63ea390d0e

.github/workflows/e2e.yml
littlewhywhat: ok, fixed here https://github.com/salsita/foosball-rating/pull/154/commits/31a8ba5da60da761aea1132031a4890dfdba22cd

.github/workflows/build.yml
littlewhywhat: and here :) https://github.com/salsita/foosball-rating/pull/154/commits/dfec6545ed6d9a98ad73fbd0ee640deb8ae7d52a

package.json
novellizator: a pity we cannot possibly make comments on this in a .json file. I'd think / should return a 200 in order to have a healthcheck implementable on it. in it's simplest form `{alive: true}` should be a sufficient response. Let's leave it for the future

PR #162, opened 2020-05-11
Implement docker-compose

novellizator:
(this is a POC, to be dropped)
Runs the foosball app from docker compose.
Goals:
 - no need to install deps like postgres into your host system
 - setup similar (almost identical) to the one on heroku - more configurable

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+*/node_modules*
+*/*/node_modules*
+**/node_modules/
+**/build
+**/.env
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,9 @@ RUN apt-get install -qq -y curl xz-utils >/dev/null
 RUN curl -s http://node.salsita.co/ | bash
 RUN node_installer 10.15.3
 
-RUN rm -rf /srv && useradd -md /srv -s /bin/bash foosball
-USER foosball
+RUN rm -rf /srv
 WORKDIR /srv
-ADD --chown=foosball:foosball . /srv/
-
+COPY . /srv/
 RUN cp frontend/.env.docker.example frontend/.env
-RUN npm install
+RUN npm install --production
 RUN npm run install:children
-RUN cd frontend && npm run build && cd ..
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,26 @@
+version: "3"
+services:
+  database:
+    image: postgres:9.6-alpine
+    container_name: foos-psql
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_DB: postgres
+  nginx-gated-app:
+    build: .
+    depends_on:
+      - "database"
+    restart: always
+    command: >
+      bash -c "npm run start:populatedb"
+    volumes:
+      - "./frontend/src:/srv/frontend/src"
+    ports:
+      - "3000:80"
+    environment:
+      NODE_ENV: production
+      DATABASE_URL: postgresql://postgres:postgres@database:5432/postgres
+      PORT: 80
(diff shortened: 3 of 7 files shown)


nginx/nginx-heroku.conf
```diff
@@ -2,10 +2,11 @@ server {
     listen %PORT%;
 
     location / {
-        root /srv/frontend/build;
-        try_files $uri /index.html;
+        # root /srv/frontend/build;
+        # try_files $uri /index.html;
         auth_basic "Restricted Content";
         auth_basic_user_file /srv/nginx/htpasswd;
+        proxy_pass http://127.0.0.1:3001/;
```
novellizator, line 9: running the develop version (npm start) not the built version

Dockerfile
```diff
@@ -5,12 +5,9 @@ RUN apt-get install -qq -y curl xz-utils >/dev/null
...
-RUN rm -rf /srv && useradd -md /srv -s /bin/bash foosball
-USER foosball
+RUN rm -rf /srv
 WORKDIR /srv
-ADD --chown=foosball:foosball . /srv/
-
+COPY . /srv/
 RUN cp frontend/.env.docker.example frontend/.env
-RUN npm install
+RUN npm install --production
```
novellizator, line 12: experiment: this shrinks the size od the container from 1.2GB to 500MB

frontend/package.json
```diff
@@ -22,8 +22,8 @@
     "url-join": "^4.0.1"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "start:e2e": "PORT=3001 BROWSER=NONE react-scripts start", 
+    "start": "PORT=3001 BROWSER=NONE react-scripts start",
```
novellizator, line 25: now these things are identical

Dockerfile
```diff
@@ -5,12 +5,9 @@ RUN apt-get install -qq -y curl xz-utils >/dev/null
...
 RUN npm run install:children
-RUN cd frontend && npm run build && cd ..
```
novellizator, line 16: in hindsight, this could have been left

docker-compose.yaml
```diff
@@ -0,0 +1,26 @@
...
+      POSTGRES_DB: postgres
+  nginx-gated-app:
+    build: .
+    depends_on:
+      - "database"
+    restart: always
+    command: >
+      bash -c "npm run start:populatedb"
+    volumes:
+      - "./frontend/src:/srv/frontend/src"
```
novellizator, line 20: this is only a covenience for developers, so after every change to the sources you don't need to recompile

package.json
```diff
@@ -22,14 +23,14 @@
   "homepage": "https://github.com/salsita/foosball-rating#readme",
   "dependencies": {
     "pg": "^7.18.1",
-    "serve": "^10.1.2"
+    "serve": "^10.1.2",
+    "concurrently": "^5.1.0"
```
novellizator, line 27: this is actually a bug on master - if you install it for production (no devDependencies), then this package is missing

.dockerignore
```diff
@@ -0,0 +1,5 @@
+*/node_modules*
+*/*/node_modules*
```
novellizator, line 2: oh, node_modules is duplicated like 3 times :D

package.json
littlewhywhat: oops...

littlewhywhat:
> no need to install deps like postgres into your host system

but you need to have docker instead of postgres and be able to run it, right? :thinking:

novellizator:
> > no need to install deps like postgres into your host system
> 
> but you need to have docker instead of postgres and be able to run it, right? 🤔

yes. but it seems like it's easier to install 1 docker than postgres + sometimes redis + sometimes nginx + i dont know what

novellizator:
This was just an experiment, did work but turning it into production qould require a rewrite

PR #165, opened 2020-05-29
Send scores

littlewhywhat:
Signed-off-by: Roman Vaivod <romanv@salsitasoft.com>


littlewhywhat:
looks like it's ready - there is even an e2e test to be sure that things work :D

PR #169, opened 2020-07-15
167: Build failures

littlewhywhat:


littlewhywhat:
so far looks like postdeploy script doesn't work at all for some reason... tried to add npm run create-schema to be sure that schema is present but it's still doesn't seem to do anything... maybe though need to reopen PR :thinking:

littlewhywhat:
Still nothing after reopen

littlewhywhat:
closed as issue was solved in #174

Issue #167, opened 2020-07-03
build failures on heroku for PRs


littlewhywhat:

littlewhywhat:
closed by 174

PR #174, opened 2020-08-10
gh-167: Fix postdeploy failures

littlewhywhat:
Just be sure that it helps - making a new PR.

Heroku support thread - https://help.heroku.com/906487

Adding ENTRYPOINT [] to Dockerfile so that our image doesn't inherit the ENTRYPOINT from the base image.


littlewhywhat:
It works! :rocket: :dancers:

PR #176, opened 2020-09-13
gh-175: Fix slackbot declaration

littlewhywhat:
* Using `export = Slackbot` instead of `export default`
* enable esModuleInterop
* have to import similarly `express`
and things start to magically work :man_shrugging:


littlewhywhat:
a bit worried about other modules that are imported as `import * as` but e2e passes and they test now basic functions... so I guess it needed to be fixed only for `moment` and `express`. What do you think? @stetizu1

